PR #1, opened 2021-10-30
Implement macro-hygiene and general improvements

daxpedda:


tests/clone.rs
```diff
@@ -6,9 +6,12 @@ use derive_restricted::derive_where;
 // #[derive_where(T: Clone; Clone)]
```
ModProg, line 6:
```suggestion
// #[derive_where(T; Clone)]
```
Maybe better make this an actual test without the generic. As Generics on Units are not supported by Rust.

src/lib.rs
```diff
@@ -1,134 +1,275 @@
...
                     }
                 }
             }
+            Data::Union(_) => todo!("Unions are not supported"),
+        };
+
+        self.generate_signature(body)
+    }
+
+    fn prepare_field_named(fields: &FieldsNamed) -> impl Iterator<Item = TokenStream> + '_ {
```
ModProg, line 180: These could move to a Trait implemented for Fields or FieldsNamed, but maybe it does make more sense to keep them here, as they would be the only two we can do that for.

tests/clone.rs
daxpedda: I think we can remove the whole thing. The crate shouldn't handle things without generics anyway, because the standard library does that already perfectly.

src/lib.rs
daxpedda: I re-worked these parts largely.

tests/clone.rs
ModProg: Yeah, we should just add a reasonable error than that tells the user to use `#[derive]`

src/lib.rs
daxpedda: These functions don't exist anymore, so I will let you re-review this.

tests/clone.rs
daxpedda: I noted this in #2.

src/lib.rs
```diff
@@ -90,173 +90,185 @@ impl TryFrom<Ident> for Traits {
...
             PartialEq => "::core::cmp::PartialEq",
             PartialOrd => "::core::cmp::PartialOrd",
-            Ord => "::core::cmp::Ord",*/
+            Ord => "::core::cmp::Ord",
         })
         .expect("couldn't pass path to trait")
     }
 
-    fn generate_body(&self, data: &Data) -> TokenStream {
+    fn generate_body(self, name: &str, data: &Data) -> TokenStream {
```
ModProg, line 108:
```suggestion
    fn generate_body(self, name: &Ident, data: &Data) -> TokenStream {
```
Is there  a reason to use a string instead?

daxpedda: Yes, the name is only used by the `Debug` implementation, it needs a string.

src/lib.rs
ModProg: Yeah, I think it is reasonable to keep them there for now, if we know for sure what we want from them, we could move them to a separate module.

ModProg:
@daxpedda I did the name:Ident I think that improves readability a bit, as the to_string actually happens where it is used.

src/lib.rs
```diff
@@ -90,173 +90,185 @@ impl TryFrom<Ident> for Traits {
...
-
-                        quote! {
-                            let Self ( #(#fields_temp),* ) = self;
-                            #body
-                        }
-                    }
-                    Fields::Unit => self.generate_unit(name),
+                    Fields::Named(fields) => self.generate_struct(name, &variant, fields),
+                    Fields::Unnamed(fields) => self.generate_tuple(name, &variant, fields),
+                    Fields::Unit => self.generate_unit(name, &variant),
```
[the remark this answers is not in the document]
daxpedda: This has been done.

src/lib.rs
daxpedda: Has been done.

daxpedda:
Let's merge this!

ModProg (approved):

Issue #2, opened 2021-10-30
Validation


daxpedda:
Currently the input on `#[derive_where(xxx)]` is not validated properly and sometimes doesn't output useful error messages.

Thinks we want to cover:
- [x] Syntax:
  - [x] Support not inputting any generics
  - [x] Check for duplicate commas
  - [x] Check for duplicate semicolons
  - [x] Pseudo-validate that generics are actual generics
  - [x] Support multiple traits to implement
- [x] Applicability
  - [x] Don't support items without generics, error message should say to use standard derives instead
  - [x] Better error message on unions, currently it's panicking
- [x] Evaluate if `proc_macro_error` can be replaced with something hand-written

daxpedda:
Research https://docs.rs/proc-macro-error/1.0.4/proc_macro_error/.

ModProg:
Well probably just use empty bounds, we could use look_ahead and try parse the DeriveWhere and if it fails parse only Trait list.
so #[derive_where(Clone)] without ; and T.

daxpedda:
We could also test for parsing `;`, if we don't find a `;`, we can assume it's only traits. We do support no generics, but traits are mandatory.

PR #13, opened 2021-10-31
Internal documentation

### LINKED ISSUE #4: Internal Documentation
macros are hard

daxpedda:
This adds internal documentation, I hope you like my style!
I also managed to remove proc-macro-error and left some TODOs where necessary.
Some code shuffling also happened.

Fixes #4.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -110,8 +110,8 @@ impl Traits {
                 let pattern = quote! { Self };
 
                 match &data.fields {
-                    Fields::Named(fields) => self.build_struct(debug_name, &pattern, None, fields),
-                    Fields::Unnamed(fields) => self.generate_tuple(debug_name, &pattern, None, fields),
+                    Fields::Named(fields) => self.build_for_struct(debug_name, &pattern, None, fields),
+                    Fields::Unnamed(fields) => self.build_for_tuple(debug_name, &pattern, None, fields),
                     fields @ Fields::Unit => return Err(Error::new(
                         fields.span(),
                         "Using `derive_where` on unit struct is not supported as unit structs don't support generics.")),
@@ -129,20 +129,20 @@ impl Traits {
                         let pattern = quote! { Self::#debug_name };
 
                         match &variant.fields {
-                            Fields::Named(fields) => self.build_struct(
+                            Fields::Named(fields) => self.build_for_struct(
                                 debug_name,
                                 &pattern,
                                 Some((index, &variants)),
                                 fields,
                             ),
-                            Fields::Unnamed(fields) => self.generate_tuple(
+                            Fields::Unnamed(fields) => self.build_for_tuple(
                                 debug_name,
                                 &pattern,
                                 Some((index, &variants)),
                                 fields,
                             ),
                             Fields::Unit => {
-                                self.generate_unit(debug_name, &pattern, Some((index, &variants)))
+                                self.build_for_unit(debug_name, &pattern, Some((index, &variants)))
                             }
                         }
                     })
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -1,49 +1,69 @@
...
 use syn::parse::{Parse, ParseStream};
 use syn::punctuated::Punctuated;
 use syn::spanned::Spanned;
 use syn::{
-    parse_macro_input, Data, DeriveInput, Error, Fields, FieldsNamed, FieldsUnnamed, Token,
-    TraitBound, Type,
+    Data, DeriveInput, Error, Fields, FieldsNamed, FieldsUnnamed, Result, Token, TraitBound, Type,
 };
 
+/// Holds parsed [bounds](TraitBound) and [traits][Traits].
```
ModProg, line 18: This looks strange, shouldn't both links have the same syntax?

src/lib.rs
```diff
@@ -1,49 +1,69 @@
...
 struct DeriveWhere {
+    /// Parsed [bounds](TraitBound).
     bounds: Vec<TraitBound>,
+    /// Parsed [traits][Traits].
```
ModProg, line 22: Same here, I think both work just interested why

src/lib.rs
```diff
@@ -447,51 +491,66 @@ impl Traits {
...
+            // If there is no `where` clause, make one.
+            let where_clause = where_clause.get_or_insert(quote! { where });
 
             for bound in &derive_where.bounds {
-                bounds.extend(quote! {
+                where_clause.extend(quote! {
                     #bound: #trait_,
                 })
             }
+        }
```
ModProg, line 530: 1. Do we need to support empty bounds?
2. I think the DeriveWhere parser fails on empty bounds or traits.

ModProg (changes requested):
https://github.com/ModProg/derive-where/blob/5712ae08e5bf06fec188e428685fbe93c1dd857f/src/lib.rs#L166
Maybe document skip, the rest is very self-explanatory 

https://github.com/ModProg/derive-where/blob/5712ae08e5bf06fec188e428685fbe93c1dd857f/src/lib.rs#L293
Maybe document pattern

src/lib.rs
daxpedda: This is a typo.

src/lib.rs
daxpedda: I copied the typo from above and didn't notice, I would also like to know why this works 😄.

src/lib.rs
daxpedda: 1. We do have to support empty bounds:
```rust
struct Test<T>(PhantomData<T>);

impl<T> Clone for Test<T> {
  fn clone(&self) -> Self {
    Self(PhantomData)
  }
}
```
2. Indeed, something we still have to support, should be tracked at #2. We still have to figure out how the syntax will look like then.

src/lib.rs

src/lib.rs
daxpedda: I did a little bit of research, using square brackets for the link should point to a reference. Like this:
```markdown
[link text][link reference]

[link reference]: https://example.com
```
But somehow, it recognizes that the reference name is a link and just uses it, I guess there is some built-in leniency in the markdown parser. I didn't dig into Rusts markdown rules.

src/lib.rs
ModProg: Well probably just use empty bounds, we could use `look_ahead` and try parse the DeriveWhere and if it fails parse only Trait list.
so `#[derive_where(Clone)]` without `;` and `T`.

ModProg (changes requested):
`build_struct` vs `generate_tuple`

document that the pattern parameter is the variant of an enum or Self we implement for

src/lib.rs
ModProg: This is part of #2 and should not block this.

daxpedda:
> `build_struct` vs `generate_tuple`

Don't know how I forgot that, I renamed them all to `build_for_x` now, I hope that's alright?
> document that the pattern parameter is the variant of an enum or Self we implement for

Right, you pointed that out above but I just forgot, sry 😅! I documented it only in `build_for_struct` and put a note in the others to look there.

ModProg (approved):

ModProg:
LGTM, anything you want to add before merging?

daxpedda:
No, looks good, I will start working on tests in a separate PR.

Issue #11, opened 2021-10-31
Publish


daxpedda:
When we finish with all our issues, we should publish the crate!

- [x] Fill `Cargo.toml` with all necessary fields
- [x] Add LICENSE files
- [x] Finish README (#12)
- [x] Decide on a name (#6)

daxpedda:
https://crates.io/crates/derive-where 🎉

PR #14, opened 2021-11-01
Add license and document missing fields in Cargo.toml

### LINKED ISSUE #6: Rename crate
Probably to `derive-where`?

ModProg:
closes #6 #11

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,7 +1,17 @@
 [package]
-name = "derive-restricted"
+name = "derive-where"
 version = "0.1.0"
 edition = "2021"
+# rust-version
+description = "Deriving with custom trait bounds"
+documentation = "https://docs.rs/derive-where"
+readme = "README.md"
+homepage = "https://github.com/ModProg/derive-where"
+repository = "https://github.com/ModProg/derive-where"
+license = "MIT OR Apache-2.0"
+# keywords
+# categories
+# include
 
 [lib]
 proc-macro = true
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,15 +1,15 @@
 [package]
 name = "derive-where"
 version = "0.1.0"
-edition = "2021"
-# rust-version
+edition = "2018"
+rust-version = "1.43.1"
 description = "Deriving with custom trait bounds"
 documentation = "https://docs.rs/derive-where"
 readme = "README.md"
 repository = "https://github.com/ModProg/derive-where"
 license = "MIT OR Apache-2.0"
-# keywords
-# categories
+keywords = ["derive", "macro"]
+categories = ["rust-patterns", "no-std"]
 include = [
 	"src/**/*",
 	"Cargo.toml",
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -1,7 +1,17 @@
 [package]
-name = "derive-restricted"
+name = "derive-where"
 version = "0.1.0"
 edition = "2021"
+# rust-version
+description = "Deriving with custom trait bounds"
+documentation = "https://docs.rs/derive-where"
+readme = "README.md"
+homepage = "https://github.com/ModProg/derive-where"
```
daxpedda, line 9: These should/can be removed:
> If no URL is specified in the manifest file, crates.io will automatically link your crate to the corresponding docs.rs page.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-documentation-field
> If no value is specified for this field, and a file named `README.md`, `README.txt` or `README` exists in the package root, then the name of that file will be used.

https://doc.rust-lang.org/cargo/reference/manifest.html#the-readme-field

We don't have a homepage, setting it to the repository is probably not helpful. The repository link is already shown on crates.io, I think this should be enough.

What do you think?

Cargo.toml
```diff
@@ -1,7 +1,17 @@
...
+repository = "https://github.com/ModProg/derive-where"
+license = "MIT OR Apache-2.0"
+# keywords
+# categories
+# include
```
daxpedda, line 14: I don't think we need the `include` tag, but we might want to use the `exclude` tag for the up and coming `.github` folder!
```suggestion
# exclude
```

daxpedda:
A couple of nits.

Cargo.toml
ModProg: I agree on the homepage url, I've seen the other fields set usually even tho they are redundant, I think it can be nice to have them set anyway, for non crates.io usage.

Cargo.toml
ModProg: yeah, I would have just added something like `clap`:
```toml
include = [
	"src/**/*",
	"Cargo.toml",
	"LICENSE-*",
	"README.md"
]
```
Examples and tests don't need to be in there as well IIRC

Cargo.toml
daxpedda: Good point!

Cargo.toml
daxpedda: For a moment I thought that's neat, better to control whats included then trying to exclude a bunch of things. But I can't find the documentation actually saying anything like that: https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields. So I think we will still need the `exclude` field, unless I'm missing something here.

I would like to know why `clap` does this, maybe they wrongly thought what I assumed here too?

ModProg: > The options are mutually exclusive; setting include will override an exclude. If you need to have exclusions to a set of include files, use the ! operator described below.

daxpedda: This doesn't tell me that using `include` will **only** include the files specified. I will try it actually.

daxpedda: I tried it, it works as you intended, LGTM.

daxpedda (approved):
LGTM, feel free to merge.

daxpedda (approved):

PR #15, opened 2021-11-01
General Improvements

### LINKED ISSUE #2: Validation

### LINKED ISSUE #3: Tests
- [ ] Integration tests
  - [ ] Tuple
    - [x] One field
    - [ ] Two fields
    - [x] Three fields
  - [ ] Struct
    - [x] One field
    - [ ] Two fields
    - [x] Three fields
  - [ ] Enum
    - [ ] One variant
      - [x] Tuple with one field
      - [x] Struct with one field
      - [ ] Tuple with two fields
      - [ ] Struct with two fields
      - [x] Tuple with three fields
      - [x] Struct with three fields
    - [ ] Two variants
      - [ ] Both variants structs
        - [ ] First struct with no fields
        - [ ] Second struct with no fields
        - [ ] Both structs with one field
        - [ ] Both structs with two fields
        - [ ] Both structs with three fields
      - [ ] Both variants tuples
        - [ ] First tuple with no fields
        - [ ] Second tuple with no fields
        - [ ] Both tuple with one field
        - [ ] Both tuple with two fields
        - [ ] Both tuple with three fields
      - [ ] First variant struct, second variant unit
        - [ ] Struct with one field
        - [ ] Struct with two fields
        - [ ] Struct with three fields
      - [ ] First variant unit, second variant struct
        - [ ] Struct with one field
        - [ ] Struct with two fields
        - [ ] Struct with three fields
      - [ ] First variant tuple, second variant unit
        - [ ] Tuple with one field
        - [ ] Tuple with two fields
        - [ ] Tuple with three fields
      - [ ] First variant unit, second variant tuple
        - [ ] Tuple with one field
        - [ ] Tuple with two fields
        - [ ] Tuple with three fields
    - [ ] Three variants
      - [ ] All variants structs
        - [ ] First variant struct with no fields
        - [ ] Second variant struct with no fields
        - [ ] Third variant struct with no fields
        - [ ] First and second variants structs with no fields
        - [ ] First and third variants structs with no fields
        - [ ] Second and third variants structs with no fields
      - [ ] First variant unit
        - [ ] All others structs
        - [ ] All others tuples
        - [ ] Second variant struct, third variant tuple
        - [ ] Second variant tuple, third variant struct
      - [ ] Second variant unit
        - [ ] All others structs
        - [ ] All others tuples
        - [ ] First variant struct, third variant tuple
        - [ ] First variant tuple, third variant struct
      - [ ] Third variant unit
        - [ ] All others structs
        - [ ] All others tuples
        - [x] First variant struct, second variant tuple
        - [ ] First variant tuple, second variant struct
      - [ ] First and second variant unit
        - [ ] Third variant struct
        - [ ] Third variant tuple
      - [ ] First and third variant unit
        - [ ] Second variant struct
        - [ ] Second variant tuple
      - [ ] Second and third variant unit
        - [ ] First variant struct
        - [ ] First variant tuple
  - [x] Edge-cases
    - [x] Associated type generics: `<T as Trait>::Type>; Clone`
    - [x] Empty bound: `Clone, Debug`
    - [x] Custom bound: `T: SomeTraitRequiringClone; Clone`
    - [x] Where clause:
      ``` rust
      struct Test<T>
      where T: SomeTrait
      {
        ...
      }
      ```
- [x] Derive tests
  - [x] Struct
  - [x] Tuple
  - [x] Enum
    - [x] Struct
    - [x] Tuple
    - [x] Unit
  - [x] Edge-cases
    - [x] No generic type parameters
    - [x] Associated type generics: `<T as Trait>::Type; Clone`
    - [x] Custom bound: `T: SomeTraitRequiringClone; Clone`
    - [x] Where clause:
      ``` rust
      struct Test<T>
      where T: SomeTrait
      {
        ...
      }
      ```
- [x] Error messages
  - [x] Union
  - [x] Item with no generics
  - [x] No trait found
  - [x] Unit struct

All these tests have to account for:
- [ ] Copy
- [ ] Clone
- [ ] Debug
- [ ] Default
- [ ] Eq
- [ ] Hash
- [ ] Ord
- [ ] PartialEq
- [ ] PartialOrd
- [ ] Zeroize

### LINKED ISSUE #5: Implement `unsafe` implementation for `Ord`
`Discriminant` can be transmuted to a `usize`. This can be used to remove the complex implementation of trying to match every variant with every variant.

This will need to be hidden behind a crate feature.

### LINKED ISSUE #7: Implement nightly implementation for `Ord`
This is related to #5.

On nightly a safe implementation can be done by using [`core::intrinsic::discriminant_value`](https://doc.rust-lang.org/core/intrinsics/fn.discriminant_value.html). This is unstable, but it can be avoided to use a feature in the top-level crate by using [`#[internal_unstable]`](https://doc.rust-lang.org/unstable-book/language-features/allow-internal-unstable.html).

This is also how `std` implements `PartialOrd` and `Ord`.

Obviously has to be hidden behind a crate feature.

### LINKED ISSUE #8: CI
Hook up CI:
- [ ] Dependabot
- [ ] cargo-audit
- [ ] Clippy
- [ ] rustfmt
- [ ] Build
  - [ ] `#[no_std]` build
  - [ ] MSRV build (#9) (https://crates.io/crates/cargo-msrv)
- [ ] Test
  - [ ] Unit and integration tests
  - [ ] Examples
  - [ ] `unsafe` feature (#5)
  - [ ] `nightly` feature (#7)

### LINKED ISSUE #10: Implement traits of external crates
We discussed implementing this for `Zeroize` for example, this might not be necessary if `Zeroize` can be fixed with a PR.
We should explore this further.

### LINKED ISSUE #18: Allow deriving `Copy` for unions
Which is the only trait supported by standard `derive`.

### LINKED ISSUE #19: `#[derive_where(skip(Clone, Debug))]` field attribute
We should add a `skip` field attribute à la zeroize.

### LINKED ISSUE #20: Auto-deref
I found the following issue, which affects `Zeroize` quiet heavily:
```rust
struct SliceDeref([u8; 1]);

impl Deref for SliceDeref {
    type Target = [u8];

    fn deref(&self) -> &Self::Target {
        &self.0
    }
}

impl DerefMut for SliceDeref {
    fn deref_mut(&mut self) -> &mut Self::Target {
        &mut self.0
    }
}

#[derive_where(Zeroize; T)]
struct Test3<T>(T, SliceDeref);

let mut test = Test3(42, SliceDeref([42]));
test.zeroize();

assert_eq!(test.0, 0);
assert_eq!(test.1 .0, [0]);
```

This will fail because `SliceDeref` doesn't implement `Zeroize` and we call `zeroize` by using the fully-qualified syntax, which prevents auto-deref.

Obviously this affects all traits, as far as I know there is no way to use FQS and auto-deref. So my current thought is to introduce another attribute that basically forces derive-where to use auto-deref instead of FQS, something like `#[derive_where(auto_deref(Clone, Zeroize))]`?

daxpedda:
I used [trybuild](https://docs.rs/trybuild) to test our error messages. Also turned out there were a lot of bugs, fixed those.

Fixes #2.
Fixes #3.
Fixes #5.
Fixes #7.
Fixes #8.
Fixes #10.
Fixes #18.
Fixes #19.
Fixes #20.


src/lib.rs
```diff
@@ -25,11 +25,28 @@ struct DeriveWhere {
...
+        let bounds = if found_semicolon {
+            let bounds = Punctuated::<TraitBound, Token![,]>::parse_separated_nonempty(input)?
+                .into_iter()
+                .collect();
+            <Token![;]>::parse(input)?;
+
+            bounds
+        } else {
+            Vec::new()
+        };
```
daxpedda, line 49: This new implementation is still up for discussion. @ModProg feel free to overwrite it.

ModProg: I would go with fork and advance_to, but let's create an issue about benchmarking this.

daxpedda: Solved in another PR.

daxpedda:
Alright, I inverted the syntax and used some custom parsing to improve error messages. This looks much better now. The only thing I would have like is improve error messages on duplicate commas, but the code got so bloated handling that, I reverted everything.

See https://github.com/ModProg/derive-where/issues/3#issuecomment-958028109.

ModProg:
> Alright, I inverted the syntax and used some custom parsing to improve error messages. This looks much better now. The only thing I would have like is improve error messages on duplicate commas, but the code got so bloated handling that, I reverted everything.

Well I don't think duplicate commas are a very common mistake, and normally easily spotted by the developer themselves

tests/ui/invalid_generic.stderr
```diff
@@ -0,0 +1,5 @@
+error: expected type to bind to, expected one of: `for`, parentheses, `fn`, `unsafe`, `extern`, identifier, `::`, `<`, square brackets, `*`, `&`, `!`, `impl`, `_`, lifetime
+ --> tests/ui/invalid_generic.rs:1:37
+  |
+1 | #[derive_where::derive_where(Clone; where)]
+  |                                     ^^^^^
```
ModProg, line 5: This looks a bit strange, we would only expect a type here, not parentheses ??

tests/ui/no_generics.rs
```diff
@@ -0,0 +1,9 @@
+#[derive_where::derive_where(Clone)]
+struct Test1(u8);
+
+// Make sure this doesn't reach unreachable code as it should fail early for
+// not having any generics.
```
ModProg, line 5: I don't disagree with erroring on these, as the standard derive is always preferable if possible, but I don't think there is unreachable code to reach

tests/ui/union.rs
```diff
@@ -0,0 +1,9 @@
+#[allow(unused_imports)]
+use core::marker::PhantomData;
+
+#[derive_where::derive_where(Clone)]
+union Test<T> {
+    field: PhantomData<T>,
+}
```
ModProg, line 7: Should we create an issue for unions? To document we could implement them if needed.

src/lib.rs
```diff
@@ -42,58 +43,63 @@ impl Parse for Generic {
...
+                    if !input.is_empty() {
+                        generics = Some(
+                            Punctuated::<Generic, Token![,]>::parse_terminated(input)?
+                                .into_iter()
+                                .collect(),
+                        );
+                    }
+                } else if let Err(error) = <Token![,]>::parse(input) {
+                    return Err(Error::new(error.span(), "expected `;` or `,"));
+                }
```
ModProg, line 98: :+1: manual parsing and yet almost more readable than my solution

src/lib.rs
```diff
@@ -554,9 +606,22 @@ impl Trait {
             Copy => quote! {},
             Debug => quote! { #pattern => ::core::fmt::Formatter::write_str(__f, #debug_name), },
             Eq => quote! {},
-            Hash => quote! { #pattern => (), },
+            Hash => {
+                // Add hashing the variant if this is an `enum`.
+                let discriminant = if variants.is_some() {
+                    let type_ = self.path();
+                    Some(quote! { #type_::hash(&::core::mem::discriminant(self), __state); })
```
ModProg, line 613: `type_` vs `path`, comparing to https://github.com/ModProg/derive-where/pull/15#diff-b1a35a68f14e696205874893c07fd24fdb88882b47c23cc0e0c80a30c7d53759R541-R546

tests/ui/invalid_generic.stderr
daxpedda: I'm not sure what you mean, I don't see any parentheses. I used `where` here as something that can't be parsed into a type.

tests/ui/no_generics.rs
daxpedda: This test is making sure that this line is never reached:
https://github.com/ModProg/derive-where/blob/d41f608e029f136e734800767d8a978ea1735c6e/src/lib.rs#L184
It should never be reached because a check happens first to check for any generics. Rust is smart enough to find no generics if they aren't used inside the item.

Maybe the comment needs to be improved?

tests/ui/union.rs
daxpedda: I briefly looked into this yesterday, just to make sure why we can't implement them. The issue is that we can't handle unions, because unlike an `enum`, we have no way to check which current variant is active. So really the only trait we could derive for unions is `Copy`, and therefore a special implementation of `Clone`, and `Hash` (I think?).

On that note, standard `derive` only support `Copy`.
So I guess yes, we should create an issue!

daxpedda: See #18.

src/lib.rs
daxpedda: Uff, the first time around when I swapped the syntax without manual parsing it was so beautiful, this looks super-messy to me now :cry:.

src/lib.rs
daxpedda: Good catch!

tests/ui/no_generics.rs
ModProg: oh ok, no it's ok. 
Even without the unreachable, we just don't want to support it anyways because it doesn't make sense. We can keep it like that.

tests/ui/invalid_generic.stderr
ModProg: I meant the `expected one of: "..."` but that is probably just the parser returning that. I think we can keep it like that as the error message pointing at the `where` and requiring a Type being pretty clear.

daxpedda:
I adjusted the MSRV to 1.36.0 because trybuild doesn't build before that and optional `dev-dependencies` aren't a thing. I also introduced `DERIVE_WHERE_SKIP_UI` environment variable, which disables UI tests, which we can use when testing for MSRV.

tests/ui/invalid_generic.stderr
daxpedda: Ah, I see. Good question, I have no clue, but yeah, this error message isn't in my control.

src/lib.rs

ModProg:
> I adjusted the MSRV to 1.36.0 because trybuild doesn't build before that

That's ok, I think 1.36 should be good enough.

If necessary, we could do the ui testing in a separate crate to not affect the MSRV. But as there is no known usecase for sub 1.36 let's keep it like it is for now.

daxpedda:
I added deriving `Copy` for unions.

daxpedda:
I added a crate feature called `safe`. My thoughts were that standard derive uses unsafe everywhere, so we should do the same by default. Switching to a safe implementation is explicit.

I also decided to make the implementation using `transmute` for the `Discriminant` for `Ord` and `PartialOrd` the default one. After digging an hour or so through the code, I determined that this has to be safe. Rust makes an explicit guarantee that single field tuple structs have to be able to be transmuted to the type of their only field. Additionally I added a test that makes sure that the inner type remains the same.

@ModProg let me know if you disagree!

ModProg:
> I added a crate feature called `safe`. My thoughts were that standard derive uses unsafe everywhere, so we should do the same by default. Switching to a safe implementation is explicit.
> 
> I also decided to make the implementation using `transmute` for the `Discriminant` for `Ord` and `PartialOrd` the default one. After digging an hour or so through the code, I determined that this has to be safe. Rust makes an explicit guarantee that single field tuple structs have to be able to be transmuted to the type of their only field. Additionally I added a test that makes sure that the inner type remains the same.
> 
> @ModProg let me know if you disagree!

No that makes definitely sense. And if we have CI running these tests we can ensure that this stays the way it is. but I also don't see any reason for them to change it.

daxpedda:
Implemented the `nightly` crate feature now. I hope the code isn't too messy.
- `safe` avoids all unsafe
- `nightly` doesn't actually use any unsafe, so it doesn't conflict with `safe`

daxpedda:
I went ahead and implemented `Zeroize`.

daxpedda:
Remaining:
- Implement `skip`
- Allow on items without generics
- Produce an error when using `default` on non-enums or without a present `Default` trait
- Finish integration tests
- Lastly, improve the code by splitting the logic and hopefully making some modules

@ModProg I know this PR has gotten rather big, it's still reviewable commit-by-commit thought (most of the times 😅). Feel free to tell me to stop and put it in another PR or review or tell me what you think.

ModProg:
> @ModProg I know this PR has gotten rather big, it's still reviewable commit-by-commit thought (most of the times sweat_smile). Feel free to tell me to stop and put it in another PR or review or tell me what you think.

It's fine, due to the extensive testing I don't think this will be a problem. 

Only thing we deffinitly need to do is split up the lib.rs 2.6k is just a bit too many lines IMO

daxpedda:
> Only thing we deffinitly need to do is split up the lib.rs 2.6k is just a bit too many lines IMO

Absolutely agreed!

ModProg:
One idea I would find interesting would be an API that allows 2 (3) different usages:

Simple case:
```rs
#[derive_where::derive_where(Copy, Clone)]
```

Different derives:
```rs
#[derive_where::derive_where]
#[derive_where(Copy, Clone)]
#[derive_where(Debug; T)]
```

Maybe useful to support:
```rs
#[derive_where::derive_where(Copy, Clone)]
#[derive_where(Debug; T)]
```

daxpedda:
This is already supported, but differently, because we switched to `proc_macro_derive`.
```rust
#[derive(derive_where::DeriveWhere)]
#[derive_where(Copy, Clone)]
#[derive_where(Debug; T)]
```

ModProg:
> This is already supported, but differently, because we switched to `proc_macro_derive`.
> 
> ```rust
> #[derive(derive_where::DeriveWhere)]
> #[derive_where(Copy, Clone)]
> #[derive_where(Debug; T)]
> ```

Yeah I know, that was more an idea of a possible api improvement that we could go for if we want to (serde_with) uses that pattern. 

But that would not change a lot internally as that would be kindof just a wrapper around everything :D

non-msrv-tests/tests/ui/no_generics.stderr
```diff
@@ -0,0 +1,15 @@
+error: derive-where doesn't support items without generics, `skip` attributes or `enum`s implementing `Default`, as this can already be handled by standard `#[derive(..)]`
+ --> tests/ui/no_generics.rs:2:1
+  |
+2 | / #[derive_where(Clone)]
+3 | | struct Test1(u8);
+  | |_________________^
+
+error: derive-where doesn't support items without generics, `skip` attributes or `enum`s implementing `Default`, as this can already be handled by standard `#[derive(..)]`
```
ModProg, line 8: We should keep in mind, that default might in the future be std: https://doc.rust-lang.org/beta/unstable-book/language-features/derive-default-enum.html

non-msrv-tests/tests/ui/not-zeroize/comma.stderr
```diff
@@ -0,0 +1,11 @@
+error: unsupported trait syntax, expected one of expected one of Clone, Copy, Debug, Default, Eq, Hash, Ord, PartialEq, PartialOrd
+ --> tests/ui/not-zeroize/comma.rs:2:22
+  |
+2 | #[derive_where(Clone,; T)]
+  |                      ^
```
ModProg, line 5: Maybe allow trailing commas? As that is the rust default.

src/data/field.rs
```diff
@@ -0,0 +1,116 @@
...
+			Member::Named(ident) => tokens.append_all(Some(ident)),
+			Member::Unnamed(index) => tokens.append_all(Some(index)),
+		}
+	}
+}
+
+impl Display for Member<'_> {
+	fn fmt(&self, f: &mut Formatter<'_>) -> fmt::Result {
+		match self {
+			Member::Named(ident) => f.write_fmt(format_args!("{}", ident)),
```
ModProg, line 51: is there a reason for not using the write! macro?

non-msrv-tests/tests/ui/no_generics.stderr
daxpedda: I saw that, we could really use `#[cfg(version(..))]`.

non-msrv-tests/tests/ui/not-zeroize/comma.stderr
daxpedda: We do allow trailing commas, but not when `;` comes afterwards. We might wanna again think about some better syntax, this is all so un-Rusty.

In any case, I can change that.

src/data/field.rs
daxpedda: I actually have no clue, I thought using `format_args` is cheaper, but now I think that's just nonsense. I will look into it.

non-msrv-tests/tests/ui/not-zeroize/comma.stderr
daxpedda: Had to bump the MSRV to 1.34 too, because 1.33 doesn't support `feature(unrestricted_attribute_tokens)`, which allows things like semi-colons in there:
https://github.com/rust-lang/rust/blob/41301c3b2371365b753c2ad6a74528a38f3815ce/compiler/rustc_feature/src/accepted.rs#L295.

daxpedda:
CI is done now too.

src/data/field.rs
daxpedda: Actually, what I wrote is exactly what `write!` does under the hood. Fixed!

non-msrv-tests/tests/ui/not-zeroize/comma.stderr
daxpedda: Done.

ModProg: yeah the syntax is indeed unrusty af :D

src/data/field.rs
ModProg: yeah I thought so thats why I was a bit confused and thought there is some msrv going on?

non-msrv-tests/tests/ui/no_generics.stderr
ModProg: yeah, we can make this an issue for now to keep this in mind

ModProg (approved):

PR #16, opened 2021-11-01
Allow deriving without trait bounds

ModProg:


daxpedda:
Re-based.

src/lib.rs
```diff
@@ -8,34 +8,57 @@ use core::cmp::Ordering;
...
 
 /// Holds parsed [bounds](TraitBound) and [traits](Traits).
-struct DeriveWhere {
-    /// Parsed [bounds](TraitBound).
-    bounds: Vec<TraitBound>,
-    /// Parsed [traits](Traits).
-    traits: Vec<Traits>,
+enum DeriveWhere {
+    /// Generic type parameters were defined.
+    WithBounds(Vec<Type>, Vec<Traits>),
```
daxpedda, line 21: I did that already in #15, we now that we have switched to the `Type` type, we shouldn't use the name `bounds` anymore, as it is wrong. Generic type parameters is the correct term, I propose using the variable name `generics`.

tests/clone.rs
```diff
@@ -1,12 +1,14 @@
 #![allow(dead_code)]
 
+use std::marker::PhantomData;
```
daxpedda, line 3:
```suggestion
use core::marker::PhantomData;
```

We should test for `#[no_std]` in the future too.

src/lib.rs
```diff
@@ -523,15 +546,10 @@ fn derive_where_internal(attr: TokenStream, item: TokenStream) -> Result<TokenSt
...
+        if let DeriveWhere::WithBounds(bounds, _) = &derive_where {
             // If there is no `where` clause, make one.
             let where_clause = where_clause.get_or_insert(quote! { where });
-
-            for bound in &derive_where.bounds {
-                where_clause.extend(quote! {
-                    #bound: #trait_,
-                })
-            }
+            *where_clause = quote! { #where_clause #(#bounds: #trait_),* };
```
daxpedda, line 552: I have a `TODO` up there saying that we shouldn't use `quote` here but actually build a proper `WhereClause`, I think we can handle it in this PR.

daxpedda:
Another thing we shortly mentioned before is adding support for custom trait bounds, maybe this would nicely fit into this PR?

ModProg:
> Another thing we shortly mentioned before is adding support for custom trait bounds, maybe this would nicely fit into this PR?

yes that makes sense, we would only support either custom bounds or standard bounds not both mixed correct?

src/lib.rs
ModProg: I'll have a look at that

daxpedda:
Considering it shouldn't be any extra effort to support mixed, why not? We should basically be able to parse
`Type[: TraitBound]`.

src/lib.rs
```diff
@@ -8,34 +8,93 @@ use core::cmp::Ordering;
...
+}
 
-        <Token![;]>::parse(input)?;
+/// Holds parsed [generics](Generic) and [traits](Traits).
+enum DeriveWhere {
+    /// Generic type parameters were defined.
+    WithGenerics(Vec<Generic>, Vec<Traits>),
+    /// Only traits were set.
+    OnlyTraits(Vec<Traits>),
+}
```
daxpedda, line 60:
```suggestion
struct DeriveWhere {
  traits: Vec<Traits>,
  generics: Option<VecGeneric>>,
```
This seems more rusty to me, what do you think?

ModProg: Don't really have a preference

src/lib.rs
```diff
@@ -77,30 +68,33 @@ impl Parse for DeriveWhere {
...
+                Ok(Self { generics, traits })
             }
         }
     }
 }
 
 /// Trait to implement.
 #[derive(Copy, Clone, Debug)]
-enum Traits {
+enum Trait {
```
daxpedda, line 97: Great change actually :+1:!

src/lib.rs
```diff
@@ -572,24 +566,44 @@ fn derive_where_internal(attr: TokenStream, item: TokenStream) -> Result<TokenSt
...
+                    Generic::CoustomBound(type_bound) => type_bound.clone(),
+                    Generic::NoBound(type_) => {
+                        let mut bounds = Punctuated::default();
+
+                        bounds.push(TypeParamBound::Trait(TraitBound {
+                            paren_token: None,
+                            modifier: syn::TraitBoundModifier::None,
+                            lifetimes: None,
+                            path: trait_.clone(),
+                        }));
```
daxpedda, line 597:
```suggestion
let bounds = Punctuated::from_iterator(Some(
  TypeParamBound::Trait(TraitBound {
    paren_token: None,
    modifier: TraitBoundModifier::None,
    lifetimes: None,
    path: trait_.clone(),
  })
));
```
To avoid unnecessary allocations.
Also import `TraitBoundModifier`.

src/lib.rs
```diff
@@ -572,24 +566,44 @@ fn derive_where_internal(attr: TokenStream, item: TokenStream) -> Result<TokenSt
...
+
+                        PredicateType {
+                            lifetimes: None,
+                            bounded_ty: type_.clone(),
+                            colon_token: Colon::default(),
+                            bounds,
+                        }
```
daxpedda, line 604:
```suggestion
syn::parse2(quote! { type_: trait_})
```
Could also work here, right?

src/lib.rs
```diff
@@ -572,24 +566,44 @@ fn derive_where_internal(attr: TokenStream, item: TokenStream) -> Result<TokenSt
...
-        // TODO: don't convert to `TokenStream`, but actually properly insert bounds.
-        let mut where_clause = where_clause.map(|where_clause| where_clause.into_token_stream());
+        let (impl_generics, type_generics, mut where_clause) = generics.split_for_impl();
+
+        let default_where = &WhereClause {
+            where_token: Where::default(),
+            predicates: Punctuated::default(),
+        };
+
+        let mut where_clause = where_clause.get_or_insert(default_where).clone();
```
daxpedda:
```suggestion
        let mut where_clause = where_clause
            .cloned()
            .get_or_insert(WhereClause {
                where_token: Where::default(),
                predicates: Punctuated::default(),
            });
```
Use `cloned` first instead of after. Feel free to use a separate variable to instantiate the `default_where`.
This whole block btw should be behind a `if !derive_where.generics.is_empty()`, because otherwise we are adding an empty `where block`.

src/lib.rs
```diff
@@ -51,22 +52,12 @@ impl Parse for Generic {
...
-    fn traits(&self) -> &[Traits] {
-        match self {
-            Self::WithGenerics(_, traits) => traits,
-            Self::OnlyTraits(traits) => traits,
-        }
-    }
+/// Holds parsed [generics](Generic) and [traits](Trait).
+struct DeriveWhere {
+    /// [generics](Generic) for where clause
+    generics: Vec<Generic>,
```
daxpedda, line 58: Maybe `Option<Vec<Generic>>`? Another allocation saved.

src/lib.rs
ModProg: should we be reparsing our output just to output it a few lines later again?

daxpedda: No agreed, this makes no sense, maybe put a comment on top saying that this whole code block is basically `syn::parse2(quote! { type_: trait_})`.

Alternatively we could just revert this whole thing, I'm in favor of leaving it though, it's better then parsing stuff.

src/lib.rs
ModProg: That would also be taken care of in: https://docs.rs/syn/1.0.81/src/syn/generics.rs.html#1297-1304

daxpedda: Yeah, I noticed that in my derive tests. But I don't think we should run unnecessary code if we don't plan to use it.

src/lib.rs
ModProg: I would expect there not being a performance problem with the previous version.
This will just be returned as a token stream, so there should be no parsing in the quote happening.

I do see your point about accuracy!

ModProg: It could make sense to move some of these in their own functions to make this more readable. But I would leave it for now, and come back later to it.

daxpedda: Sounds good.

src/lib.rs
```diff
@@ -513,25 +566,44 @@ fn derive_where_internal(attr: TokenStream, item: TokenStream) -> Result<TokenSt
...
+    for trait_ in derive_where.traits {
         let body = trait_.generate_body(&ident, &data)?;
-        let trait_ = trait_.type_();
+        let trait_ = trait_.path();
 
         // Build necessary generics to construct the implementation item.
-        let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
-        // TODO: don't convert to `TokenStream`, but actually properly insert bounds.
-        let mut where_clause = where_clause.map(|where_clause| where_clause.into_token_stream());
+        let (impl_generics, type_generics, mut where_clause) = generics.split_for_impl();
```
ModProg, line 574: Is there any situation in which these can contain anything?

I don't think the DeriveInput can have a where_clause :thinking:

daxpedda: I was surprised myself when I found out that they can:
https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=bc8b601e367637bc848f0c43e100958a

ModProg: You learn something new every day.

src/lib.rs
```diff
@@ -571,39 +573,39 @@ fn derive_where_internal(attr: TokenStream, item: TokenStream) -> Result<TokenSt
...
-
-                        bounds.push(TypeParamBound::Trait(TraitBound {
-                            paren_token: None,
-                            modifier: syn::TraitBoundModifier::None,
-                            lifetimes: None,
-                            path: trait_.clone(),
-                        }));
-
-                        PredicateType {
+        let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
```
daxpedda, line 576: Considering we are cloning the `were_clause` in the for-loop now every time, which makes sense, I think this line can be moved outside the for-loop.

daxpedda (approved):

PR #31, opened 2022-01-04
Add readme-sync

daxpedda:
This makes sure our `README` and documentation in `lib.rs` are in sync. I also added a `WRITE_README=1` environment variable to the test, so we can just update the `README` instead of manually syncing them by hand.

Remaining issues:
- Currently links are a bit broken, see zheland/readme-sync#1, the workaround currently requires to specify links in `lib.rs` that aren't really necessary.
- Currently writing the `README` doesn't produce perfect output, see zheland/readme-sync#1.


ModProg:
Running `cargo test readme_sync` twice with `WRITE_README=1` fails on: 
```
`Text("([#27]):")`
 does not match docs node
`Text("(")`
  --> README.md:12:1
   |
12 | ([\#27]):
   | ^^ ^^^^^^
   |
  ::: src/lib.rs/parsed
   |
10 |  ([#27]):
   |  ^
   |
  ::: src/lib.rs
   |
16 | //! ([#27]):
   | ------------
   ```

File: non-msrv-tests/src/lib.rs [hunk 1]
```diff
@@ -10,3 +10,35 @@ fn ui() {
...
+	match std::env::var("WRITE_README") {
+		Ok(value) if value == "1" => {
+			let mut file = String::new();
+			pulldown_cmark_to_cmark::cmark(docs.iter_events(), &mut file, None).unwrap();
+			std::fs::write("../README.md", file).unwrap();
+		}
+		_ => (),
+	}
+
+	assert_sync(&readme, &docs);
```
ModProg, line 43: This does not take into effect the changes to the readme it does itself with `"WRITE_README"`

ModProg (changes requested):

non-msrv-tests/src/lib.rs
daxpedda: This is what I meant above (I had a wrong link, corrected it, see zheland/readme-sync#2). Consider this broken for now.

daxpedda:
We will see how the issues I linked above get resolved, but I don't personally have time to do it myself.

I consider the linking issue blocking, the issue with weird output when writing can be ignored, the whole writing part can be removed, checking if both are in sync is still useful and works correctly.

PR #32, opened 2022-01-05
Implement `ZeroizeOnDrop`

daxpedda:
This is waiting for RustCrypto/utils#700 and a new release. Deprecates `Zeroize(drop)` in favor of `ZeroizeOnDrop`.

### DIFF
--- a/non-msrv-tests/tests/ui/zeroize/zeroize.rs
+++ b/non-msrv-tests/tests/ui/zeroize/zeroize.rs
@@ -25,27 +25,7 @@ struct WrongCrateSyntax<T>(PhantomData<T>);
 struct InvalidCrate<T>(PhantomData<T>);
 
 #[derive(DeriveWhere)]
-#[derive_where(Zeroize(drop, drop))]
-struct DuplicateDrop<T>(PhantomData<T>);
-
-#[derive(DeriveWhere)]
 #[derive_where(Zeroize(crate = "zeroize_", crate = "zeroize_"))]
 struct DuplicateCrate<T>(PhantomData<T>);
-
-#[derive(DeriveWhere)]
-#[derive_where(Zeroize(drop, drop, crate = "zeroize_"))]
-struct DuplicateDropWithCrate<T>(PhantomData<T>);
-
-#[derive(DeriveWhere)]
-#[derive_where(Zeroize(drop, crate = "zeroize_", crate = "zeroize_"))]
-struct DropWithDuplicateCrate<T>(PhantomData<T>);
-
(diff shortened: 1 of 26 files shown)


non-msrv-tests/tests/ui/zeroize/zeroize.rs
```diff
@@ -24,28 +24,8 @@ struct WrongCrateSyntax<T>(PhantomData<T>);
 #[derive_where(Zeroize(crate = "struct Test"))]
 struct InvalidCrate<T>(PhantomData<T>);
 
-#[derive(DeriveWhere)]
-#[derive_where(Zeroize(drop, drop))]
-struct DuplicateDrop<T>(PhantomData<T>);
-
```
ModProg, line 30: maybe an error when you still have a drop telling you to use ZeroizeOnDrop instead

src/input.rs
```diff
@@ -144,10 +144,11 @@ impl<'a> Input<'a> {
...
-						crate_: Some(_), ..
-					} = **trait_
-					{
+					if let DeriveTrait::Zeroize { crate_: Some(_) } = **trait_ {
+						continue;
+					}
+					// `ZeroizeOnDrop(crate = "..")` is used.
+					if let DeriveTrait::ZeroizeOnDrop { crate_: Some(_) } = **trait_ {
 						continue;
 					}
```
ModProg, line 153: wouldn't that fit in a single match statement `Zeroize | ZeroizeOnDrop` ?

ModProg (changes requested):

non-msrv-tests/tests/ui/zeroize/zeroize.rs
daxpedda: Deprecation notice, good idea, on it.

src/input.rs
daxpedda: Indeed! On it.

non-msrv-tests/tests/ui/zeroize/zeroize.rs
daxpedda: Done.

src/input.rs
daxpedda: Turned out to be not much prettier, used an `if` because I only needed one pattern.

daxpedda:
This is ready to be merged now that the new version of `Zeroize` is out. I'm still providing backwards compatibility, especially because the new version is not a stable release.

I would say we can talk about actually enabling this feature by default at some point in the future.

ModProg (approved):

PR #33, opened 2022-01-05
derive_where attribute macro instead of derive macro

### LINKED ISSUE #27: Remove need for derive
This could easily be done if rust-lang/rust#65823 was implemented.

ModProg:
closes #27

### DIFF
--- a/.config/topic.dic
+++ b/.config/topic.dic
@@ -3,11 +3,9 @@ Changelog
 CHANGELOG
 destructure/G
 enum/S
-FIXME
 FQS
 io
 MSRV
-proc_macro_attribute
 std's
 struct/S
 TODO
--- a/.config/topic.dic
+++ b/.config/topic.dic
@@ -3,9 +3,11 @@ Changelog
 CHANGELOG
 destructure/G
 enum/S
+FIXME
 FQS
 io
 MSRV
+proc_macro_attribute
 std's
 struct/S
 TODO
--- a/README.md
+++ b/README.md
@@ -16,7 +16,6 @@ statements, with the only caveat that it requires to derive `DeriveWhere`
 ([#27]):
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Clone, Debug)]
 struct Example<T>(PhantomData<T>);
 ```
@@ -34,7 +33,6 @@ specified. This example will restrict the implementation for `Example` to
 `T: Clone`:
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Clone; T)]
 struct Example<T, U>(T, PhantomData<U>);
 ```
@@ -45,7 +43,6 @@ bind implementation for `Example` to `T: Super`:
 ```rust
 trait Super: Clone {}
 
-#[derive(DeriveWhere)]
 #[derive_where(Clone; T: Super)]
 struct Example<T>(PhantomData<T>);
 ```
@@ -65,7 +62,6 @@ impl Trait for Impl {
 	type Type = i32;
 }
 
-#[derive(DeriveWhere)]
 #[derive_where(Clone; T::Type)]
 struct Example<T: Trait>(T::Type);
 ```
@@ -75,7 +71,6 @@ specific constrain. It is also possible to use multiple separate
 constrain specifications when required:
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Clone; T)]
 #[derive_where(Debug; U)]
 struct Example<T, U>(PhantomData<T>, PhantomData<U>);
@@ -87,7 +82,6 @@ Deriving [`Default`] on an enum is not possible in Rust at the moment.
 Derive-where allows this with a `default` attribute:
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Default)]
 enum Example<T> {
 	#[derive_where(default)]
@@ -102,7 +96,6 @@ that allow it, which are: [`Debug`], [`Hash`], [`Ord`](https://doc.rust-lang.org
 [`PartialEq`](https://doc.rust-lang.org/core/cmp/trait.PartialEq.html), [`Zeroize`] and [`ZeroizeOnDrop`].
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Debug, PartialEq; T)]
 struct Example<T>(#[derive_where(skip)] T);
 
@@ -113,14 +106,12 @@ assert_eq!(Example(42), Example(0));
 It is also possible to skip all fields in an item or variant if desired:
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Debug)]
 #[derive_where(skip_inner)]
 struct StructExample<T>(T);
 
 assert_eq!(format!("{:?}", StructExample(42)), "StructExample");
 
-#[derive(DeriveWhere)]
 #[derive_where(Debug)]
 enum EnumExample<T> {
 	#[derive_where(skip_inner)]
@@ -134,7 +125,6 @@ Selective skipping of fields for certain traits is also an option, both in
 `skip` and `skip_inner`:
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Debug, PartialEq)]
 #[derive_where(skip_inner(Debug))]
 struct Example<T>(i32, PhantomData<T>);
@@ -156,7 +146,6 @@ assert_ne!(
   This is to avoid ambiguity between another method also called `zeroize`.
 
 ```rust
-#[derive(DeriveWhere)]
 #[derive_where(Zeroize(crate = "zeroize_"))]
 struct Example(#[derive_where(Zeroize(fqs))] i32);
 
@@ -189,8 +178,7 @@ and can be implemented without [`Zeroize`], otherwise it only implements
 - `crate`: an item-level option which specifies a path to the `zeroize`
   crate in case of a re-export or rename.
 
-```
-#[derive(DeriveWhere)]
+```rust
 #[derive_where(ZeroizeOnDrop(crate = "zeroize_"))]
 struct Example(i32);
 
--- a/README.md
+++ b/README.md
@@ -240,8 +240,8 @@ accompanied by a minor version bump. If MSRV is important to you, use
 
 [derivative](https://crates.io/crates/derivative)
 ([![Crates.io](https://img.shields.io/crates/v/derivative.svg)](https://crates.io/crates/derivative))
-is a great alternative with many options. Notably it has no `no_std`
-support and requires to `#[derive(Derivative)]` to use.
+is a great alternative with many options. Notably it doesn't support `no_std`
+and requires an extra `#[derive(Derivative)]` to use.
 
 ## Changelog
 
--- a/README.md
+++ b/README.md
@@ -12,8 +12,7 @@ generic type bounds.
 ## Usage
 
 The `derive_where` macro can be used just like std's `#[derive(...)]`
-statements, with the only caveat that it requires to derive `DeriveWhere`
-([#27]):
+statements:
 
 ```rust
 #[derive_where(Clone, Debug)]
@@ -242,7 +241,7 @@ accompanied by a minor version bump. If MSRV is important to you, use
 [derivative](https://crates.io/crates/derivative)
 ([![Crates.io](https://img.shields.io/crates/v/derivative.svg)](https://crates.io/crates/derivative))
 is a great alternative with many options. Notably it has no `no_std`
-support.
+support and requires to `#[derive(Derivative)]` to use.
 
 ## Changelog
 
--- a/non-msrv-tests/tests/ui/default.rs
+++ b/non-msrv-tests/tests/ui/default.rs
@@ -1,25 +1,21 @@
 use std::marker::PhantomData;
 
-use derive_where::DeriveWhere;
+use derive_where::derive_where;
 
-#[derive(DeriveWhere)]
 #[derive_where(Debug)]
 struct DefaultOnStruct<T>(#[derive_where(default)] PhantomData<T>);
 
-#[derive(DeriveWhere)]
 #[derive_where(Clone)]
 enum DefaultWithoutTrait<T> {
 	#[derive_where(default)]
 	A(PhantomData<T>),
 }
 
-#[derive(DeriveWhere)]
 #[derive_where(Default)]
 enum MissingDefault<T> {
 	A(PhantomData<T>),
 }
 
-#[derive(DeriveWhere)]
 #[derive_where(Default)]
 enum DuplicateDefaultSeparate<T> {
 	#[derive_where(default)]
@@ -28,14 +24,12 @@ enum DuplicateDefaultSeparate<T> {
 	B(PhantomData<T>),
 }
 
-#[derive(DeriveWhere)]
 #[derive_where(Default)]
 enum DuplicateDefaultSame<T> {
 	#[derive_where(default, default)]
 	A(PhantomData<T>),
 }
 
-#[derive(DeriveWhere)]
 #[derive_where(Default)]
 enum DuplicateDefaultSameSeparate<T> {
 	#[derive_where(default)]
--- a/non-msrv-tests/tests/ui/default.rs
+++ b/non-msrv-tests/tests/ui/default.rs
@@ -1,3 +1,5 @@
+use std::marker::PhantomData;
+
 use derive_where::derive_where;
 
 #[derive_where(Debug)]
--- a/non-msrv-tests/tests/ui/default.rs
+++ b/non-msrv-tests/tests/ui/default.rs
@@ -1,5 +1,3 @@
-use std::marker::PhantomData;
-
 use derive_where::derive_where;
 
 #[derive_where(Debug)]
--- a/non-msrv-tests/tests/ui/default.stderr
+++ b/non-msrv-tests/tests/ui/default.stderr
@@ -1,37 +1,37 @@
 error: unknown option
- --> tests/ui/default.rs:4:42
+ --> tests/ui/default.rs:6:42
   |
-4 | struct DefaultOnStruct<T>(#[derive_where(default)] PhantomData<T>);
+6 | struct DefaultOnStruct<T>(#[derive_where(default)] PhantomData<T>);
   |                                          ^^^^^^^
 
 error: `default` is only supported if `Default` is being implemented
- --> tests/ui/default.rs:8:17
-  |
-8 |     #[derive_where(default)]
-  |                    ^^^^^^^
+  --> tests/ui/default.rs:10:17
+   |
+10 |     #[derive_where(default)]
+   |                    ^^^^^^^
 
 error: required `default` option on a variant if `Default` is being implemented
-  --> tests/ui/default.rs:12:1
+  --> tests/ui/default.rs:14:1
    |
-12 | #[derive_where(Default)]
+14 | #[derive_where(Default)]
    | ^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = note: this error originates in the attribute macro `derive_where` (in Nightly builds, run with -Z macro-backtrace for more info)
 
 error: multiple `default` options in enum
-  --> tests/ui/default.rs:21:17
+  --> tests/ui/default.rs:23:17
    |
-21 |     #[derive_where(default)]
+23 |     #[derive_where(default)]
    |                    ^^^^^^^
 
 error: duplicate `default` option
-  --> tests/ui/default.rs:27:26
+  --> tests/ui/default.rs:29:26
    |
-27 |     #[derive_where(default, default)]
+29 |     #[derive_where(default, default)]
    |                             ^^^^^^^
 
 error: duplicate `default` option
-  --> tests/ui/default.rs:34:17
+  --> tests/ui/default.rs:36:17
    |
-34 |     #[derive_where(default)]
+36 |     #[derive_where(default)]
    |                    ^^^^^^^
--- a/non-msrv-tests/tests/ui/default.stderr
+++ b/non-msrv-tests/tests/ui/default.stderr
@@ -1,38 +1,37 @@
 error: unknown option
- --> tests/ui/default.rs:7:42
+ --> tests/ui/default.rs:4:42
   |
-7 | struct DefaultOnStruct<T>(#[derive_where(default)] PhantomData<T>);
+4 | struct DefaultOnStruct<T>(#[derive_where(default)] PhantomData<T>);
   |                                          ^^^^^^^
 
 error: `default` is only supported if `Default` is being implemented
-  --> tests/ui/default.rs:12:17
-   |
-12 |     #[derive_where(default)]
-   |                    ^^^^^^^
+ --> tests/ui/default.rs:8:17
+  |
+8 |     #[derive_where(default)]
+  |                    ^^^^^^^
 
 error: required `default` option on a variant if `Default` is being implemented
-  --> tests/ui/default.rs:17:1
+  --> tests/ui/default.rs:12:1
    |
-17 | / #[derive_where(Default)]
-18 | | enum MissingDefault<T> {
-19 | |     A(PhantomData<T>),
-20 | | }
-   | |_^
+12 | #[derive_where(Default)]
+   | ^^^^^^^^^^^^^^^^^^^^^^^^
+   |
+   = note: this error originates in the attribute macro `derive_where` (in Nightly builds, run with -Z macro-backtrace for more info)
 
 error: multiple `default` options in enum
-  --> tests/ui/default.rs:27:17
+  --> tests/ui/default.rs:21:17
    |
-27 |     #[derive_where(default)]
+21 |     #[derive_where(default)]
    |                    ^^^^^^^
 
 error: duplicate `default` option
-  --> tests/ui/default.rs:34:26
+  --> tests/ui/default.rs:27:26
    |
-34 |     #[derive_where(default, default)]
+27 |     #[derive_where(default, default)]
    |                             ^^^^^^^
 
 error: duplicate `default` option
-  --> tests/ui/default.rs:42:17
+  --> tests/ui/default.rs:34:17
    |
-42 |     #[derive_where(default)]
+34 |     #[derive_where(default)]
    |                    ^^^^^^^
--- a/non-msrv-tests/tests/ui/default.stderr
+++ b/non-msrv-tests/tests/ui/default.stderr
@@ -11,12 +11,12 @@ error: `default` is only supported if `Default` is being implemented
    |                    ^^^^^^^
 
 error: required `default` option on a variant if `Default` is being implemented
-  --> tests/ui/default.rs:14:1
+  --> tests/ui/default.rs:15:1
    |
-14 | #[derive_where(Default)]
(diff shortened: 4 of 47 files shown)


daxpedda:
Lol, after all this talk on how to do this, it looks ingenius to me. Looking forward to see how this plays out.

ModProg:
> Lol, after all this talk on how to do this, it looks ingenius to me. Looking forward to see how this plays out.

The only problem seams to be that prior to `1.52.1`, you cannot have an attribute_macro be the same identifier as a derive_helper:
```
error[E0659]: `derive_where` is ambiguous (derive helper attribute vs any other name)
  --> tests/bound.rs:21:2
   |
21 |     #[crate::derive_where(Clone; T::Type)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ ambiguous name
   |
note: `derive_where` could refer to the derive helper attribute defined here
  --> tests/bound.rs:21:2
   |
21 |     #[crate::derive_where(Clone; T::Type)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
note: `derive_where` could also refer to the attribute macro imported here
  --> tests/bound.rs:5:20
   |
5  | use derive_where::{derive_where, DeriveWhere};
   |                    ^^^^^^^^^^^^
   = help: use `crate::derive_where` to refer to this attribute macro unambiguously
```
   
Only way around this is removing the import for the derive_where macro and calling it like so:
```rs
#[derive_where::derive_where(Clone; T::Type)]
struct Test<T: Trait>(T::Type);
```

This can be somewhat circumvented by using an internal replacement: `#[__derive_where_internal_derive_attr(#attr)]
`.
But that breaks down as soon as you add additional `derive_where` attributes for `default` or `skip` or something else.

So pre `1.52.1` do not import `derive_where::derive_where`, use an alias or use the full path when invoking the macro:
```rs
use derive_where::derive_where as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[derive_where(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[derive_where(default)]
	A {
		a: Wrapper<T>,
	},
	B(Wrapper<T>),
	C,
}
```
But since `1.52.1` this would enable using it like so:
```rs
use derive_where::derive_where;
#[derive_where(Clone, Copy, Debug)]
#[derive_where(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[derive_where(default)]
	A {
		a: Wrapper<T>,
	},
	B(Wrapper<T>),
	C,
}
```

daxpedda:
I'm starting to remember what the issues were, how are we going to resolve multiple calls and paths?
```rust
use derive_where::derive_where as blubb;

#[blubb(Clone, Copy, Debug)]
#[blubb(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
struct Test<T: Trait>(T::Type);
```

ModProg:
> I'm starting to remember what the issues were, how are we going to resolve multiple calls and paths?

I would rather have to do 
```
use derive_where::derive_where as blubb;

#[blubb(Clone, Copy, Debug)]
#[derive_where(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
struct Test<T: Trait>(T::Type);
```
and if I only need one be able to do:

```
use derive_where::derive_where as blubb;

#[blubb(Clone, Copy, Debug)]
struct Test<T: Trait>(T::Type);
```

than always have the `#[derive(DeriveWhere)]`

we can make this a feature that can be opt-out/opt-in, if there is no way around this limitation.

another problem is, that in this case the first invocation would fail on the `default`
```rs
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[derive_where(default)]
	A,
    B(PhantomData<T>),
}
```

But with the internal_derive I can fix multiple `dw_macro` calls. but not multiple `dw_macro` calls plus a `derive_where` attr:
```rs
#[dw_macro(Clone, Copy)]
#[dw_macro(Eq, Hash, Ord, PartialEq, PartialOrd)]
#[derive_where(Debug)]
enum Test<T> {
	A {
		a: Wrapper<T>,
	},
	B(Wrapper<T>),
	C,
}
``` 
But this would have a msrv of `1.57.0`:
```rs
use derive_where::derive_where as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(Eq, Hash, Ord, PartialEq, PartialOrd)]
```
as prior you will get
```
error: macro attributes must be placed before `#[derive]`
  --> tests/enum_mixed.rs:18:2
   |
18 |     #[dw_macro(Eq, Hash, Ord, PartialEq, PartialOrd)]
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```





## TLDR
We can support prior `1.52.1`:
```rs
use derive_where::derive_where as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[derive_where(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[derive_where(default)]
	A,
	B(Wrapper<T>),
}
```

We can support prior `1.57.0`:
```rs
use derive_where::derive_where;
#[derive_where(Clone, Copy, Debug)]
#[derive_where(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[derive_where(default)]
	A,
	B(Wrapper<T>),
}
```

With current Rust, we can support:
```rs
use derive_where::derive_where  as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[derive_where(default)]
	A,
	B(Wrapper<T>),
}
```

But probably neither:
```rs
use derive_where::derive_where  as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(, Eq, Hash, Ord, PartialEq, PartialOrd)]
#[derive_where(Default)]
enum Test<T> {
	#[derive_where(default)]
	A,
	B(Wrapper<T>),
}
```
nor:
```rs
use derive_where::derive_where  as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[dw_macro(default)]
	A,
	B(Wrapper<T>),
}
```

ModProg:
Ok maybe we can do in current rust
```rs
use derive_where::derive_where  as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(, Eq, Hash, Ord, PartialEq, PartialOrd)]
#[derive_where(Default)]
enum Test<T> {
	#[derive_where(default)]
	A,
	B(Wrapper<T>),
}
```

The idea would be to replace all `derive_where` calls with
```rs
#[derive(DeriveWhere)]
#[__derive_where_internal_derive_attr]
```

ModProg:
If rust-lang/rust#53012 is fixed, we could also do this:
```rs
use derive_where::derive_where  as dw_macro;
#[dw_macro(Clone, Copy, Debug)]
#[dw_macro(Default, Eq, Hash, Ord, PartialEq, PartialOrd)]
enum Test<T> {
	#[dw_macro(default)]
	A,
	B(Wrapper<T>),
}
```

All we need to do is have the attribute macro expand to `derive_where` when on a field.
for now the only way would be to pass the attribute name to `dw_macro` and have the first procmacro do the replacing. But that feals so unergonomic that I would just say *If you want to use attributes on fields, you have to use `derive_where` even when the macro has a different name*

ModProg:
Oh and we would need to throw out all the sanity checks like *dont have `default` if there is no `derive_where(Default)`* as no single invocation knows which other invocation will contain `Default`.

ModProg:
As sanity checks only need to be disabled if someone renames `derive_where` and calls the macro multiple times instead of using the derive attribute `derive_where` I think making this usage an opt-in feature to enable sounds like a good idea, (If we even want to support it) by having an `disable_sanity_checks` attribute/feature.

ModProg:
As the usage of the attribute is pretty seamless (as far as I can tell) in up-to-date rust versions, we can make this the normal (maybe even only) way of using `derive_where`. 
But as we are using the `#[derive(DeriveWhere)]` way internally, we could also keep it as a different (maybe deprecated) way of doing it.

non-msrv-tests/tests/ui/item.stderr
```diff
@@ -1,53 +1,51 @@
-error: no traits found to implement, use `#[derive_where(..)` to specify some
- --> tests/ui/item.rs:6:1
+error: empty `derive_where` found
+ --> tests/ui/item.rs:7:1
```
ModProg, line 2: This message changed for some reason

non-msrv-tests/tests/ui/item_option_syntax.stderr
```diff
@@ -1,41 +1,49 @@
...
+ --> tests/ui/item_option_syntax.rs:4:18
+  |
+4 | #[derive_where = invalid]
+  |                  ^^^^^^^
+
+error: key-value macro attributes are not supported
+ --> tests/ui/item_option_syntax.rs:7:1
+  |
+7 | #[derive_where = "invalid"]
+  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ModProg, line 17: This extensive error leads to the test failing either for zeroize for no-zeroize

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
...
+				.for_each(|Variant { attrs, fields, .. }| {
+					remove_derive_where(attrs);
+					remove_derive_where_from_fields(fields)
+				})
+		}
+		Data::Union(DataUnion { fields, .. }) => remove_derive_where_from_fields_named(fields),
+	}
+
+	input.to_token_stream()
+}
```
ModProg, line 547: Should this move in a separate module?

src/test/mod.rs
```diff
@@ -12,8 +12,11 @@ use quote::quote;
 use syn::Result;
 
 fn test_derive(input: TokenStream, expected: TokenStream) -> Result<()> {
-	let left = crate::derive_where_internal(input)?.to_string();
-	let right = quote! { #expected }.to_string();
+	let left = crate::derive_where_internal(input.clone())?.to_string();
+	// FIXME this currently means, that `input_without_derive_where_attributes` is
+	// not actually tested.
+	let item = crate::input_without_derive_where_attributes(syn::parse2(input)?);
+	let right = quote! {#item #expected}.to_string();
```
ModProg, line 19: We would need to change every unit test to include the item without derive_where attributes to remove this.

src/lib.rs
daxpedda: I vote for keeping it here. I'm fine either way though.

README.md
```diff
@@ -254,7 +241,7 @@ accompanied by a minor version bump. If MSRV is important to you, use
 [derivative](https://crates.io/crates/derivative)
 ([![Crates.io](https://img.shields.io/crates/v/derivative.svg)](https://crates.io/crates/derivative))
 is a great alternative with many options. Notably it has no `no_std`
-support.
+support and requires to `#[derive(Derivative)]` to use.
```
daxpedda, line 244:
```suggestion
is a great alternative with many options. Notably it doesn't support `no_std`
and requires an extra `#[derive(Derivative)]` to use.
```

While are here, too bad none of us are native speakers :smile:.

non-msrv-tests/tests/ui/default.rs
```diff
@@ -1,25 +1,19 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 1: How did you get away with removing `use std::marker::PhantomData;`.

non-msrv-tests/tests/ui/item.stderr
daxpedda: Yes, there a test case that needs to be removed actually.

non-msrv-tests/tests/ui/item.rs
```diff
@@ -1,39 +1,30 @@
 use std::marker::PhantomData;
 
-use derive_where::DeriveWhere;
+use derive_where::derive_where;
 
-#[derive(DeriveWhere)]
 struct NoAttribute<T>(PhantomData<T>);
```
daxpedda, line 5: This test case doesn't make any sense anymore.
```suggestion
```

non-msrv-tests/tests/ui/item.stderr
daxpedda: See https://github.com/ModProg/derive-where/pull/33/files#r780830161.

non-msrv-tests/tests/ui/item_option_syntax.rs
```diff
@@ -1,29 +1,21 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
 
-use derive_where::DeriveWhere;
-
-#[derive(DeriveWhere)]
 // Rust itself already fails to parse this and will provide a separate error message.
 #[derive_where = invalid]
 struct InvalidAttribute<T>(PhantomData<T>);
```
daxpedda, line 5: This test-case can be removed now.

non-msrv-tests/tests/ui/item_option_syntax.stderr
daxpedda: See https://github.com/ModProg/derive-where/pull/33/files#r780830288.

non-msrv-tests/tests/ui/item_skip.rs
```diff
@@ -1,59 +1,47 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 1: Lol, again, how did you get away with deleting `use std::marker::PhantomData;`, magic.

non-msrv-tests/tests/ui/not-zeroize/item.rs
```diff
@@ -1,28 +1,20 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 1: Again ... missing `use std::marker::PhantomData;`.

non-msrv-tests/tests/ui/option.rs
```diff
@@ -1,34 +1,26 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 1: Missing `use std::marker::PhantomData;`.

non-msrv-tests/tests/ui/skip.rs
```diff
@@ -1,60 +1,48 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 1: Missing `use std::marker::PhantomData;`

non-msrv-tests/tests/ui/zeroize/deprecated_drop.rs
```diff
@@ -1,10 +1,7 @@
 extern crate zeroize_ as zeroize;
 
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 3: Missing `use std::marker::PhantomData;`.

non-msrv-tests/tests/ui/zeroize/item.rs
```diff
@@ -1,28 +1,20 @@
-use std::marker::PhantomData;
+use derive_where::derive_where;
```
daxpedda, line 1: Missing `use std::marker::PhantomData;`.

src/lib.rs
```diff
@@ -314,7 +301,7 @@
 //! [derivative](https://crates.io/crates/derivative)
 //! ([![Crates.io](https://img.shields.io/crates/v/derivative.svg)](https://crates.io/crates/derivative))
 //! is a great alternative with many options. Notably it has no `no_std`
-//! support.
+//! support and requires to `#[derive(Derivative)]` to use.
```
daxpedda, line 304: Sync with README there.

src/lib.rs
```diff
@@ -423,17 +423,21 @@ fn derive_where_internal(input: TokenStream) -> Result<TokenStream> {
 	// Save `Span` before we consume `input` when parsing it.
 	let span = input.span();
 	let item = syn::parse2::<DeriveInput>(input).expect("derive on unparsable item");
+	let cleaned_item = input_without_derive_where_attributes(item.clone());
```
daxpedda, line 426: I see now how you were able to skip importing `PhantomData`. The issue here is a bit complicated. When the attribute proc macro fails for any reason, the item also gets removed. This has a very bad user experience, because a missing item can break everything, instead of showing a useful error.

- Save input.
- Parse input, on failure, output both the error and the unparsed input.
- Clean input.
- Do our thing, on failure, output both the error and the cleaned input.

I suggest we handle all of this in `derive_where()`, cleaning the item has also be moved there. This will require some pretty big `clone()`s.

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
 		}
 	}
 }
+
+/// Removes the `derive_where` attributes from all fields, and variants.
+///
+/// This is needed, because rust does not support those for proc_macro_attribute
+/// currently
```
daxpedda, line 501:
```suggestion
/// Removes `derive_where` attributes from the item and all fields and variants.
///
/// This is necessary because Rust currently does not support helper attributes for attribute proc-macros and therefore doesn't automatically remove them.
```

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
...
+fn input_without_derive_where_attributes(mut input: DeriveInput) -> TokenStream {
+	use syn::Data;
+	let DeriveInput { data, attrs, .. } = &mut input;
+
+	/// Remove all attributes that are `derive_where`
```
daxpedda, line 506:
```suggestion
	/// Remove all `derive_where` attributes.
```

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
...
+	fn remove_derive_where(attrs: &mut Vec<Attribute>) {
+		// TODO find the actual path
```
daxpedda, line 508: This is impossible, unless Rust does proc macros v3 or something, lol.
```suggestion
```

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
...
+		let ident = Ident::new("derive_where", Span::call_site());
+		attrs.retain(|attr| !attr.path.is_ident(&ident))
+	}
+
+	/// Remove all attributes that are `derive_where` from `FieldsNamed`
```
daxpedda, line 513:
```suggestion
	/// Remove all `derive_where` attributes from [`FieldsNamed`].
```

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
...
+	fn remove_derive_where_from_fields_named(fields: &mut FieldsNamed) {
+		let FieldsNamed { named, .. } = fields;
+		named
+			.iter_mut()
+			.for_each(|field| remove_derive_where(&mut field.attrs))
+	}
+
+	/// Remove all attributes that are `derive_where` from `Fields`
```
daxpedda, line 521:
```suggestion
	/// Remove all `derive_where` attributes from [`Fields`].
```

src/lib.rs
```diff
@@ -490,3 +494,54 @@ fn generate_body(trait_: &DeriveTrait, item: &Item) -> TokenStream {
...
+		match fields {
+			Fields::Named(fields) => remove_derive_where_from_fields_named(fields),
+			Fields::Unnamed(FieldsUnnamed { unnamed, .. }) => unnamed
+				.iter_mut()
+				.for_each(|field| remove_derive_where(&mut field.attrs)),
+			Fields::Unit => (),
+		}
+	}
+
+	remove_derive_where(attrs);
```
daxpedda, line 532:
```suggestion
	// Remove `derive_where` attributes from the item.
	remove_derive_where(attrs);
```

src/test/mod.rs
daxpedda: I'm not following, can you explain?

daxpedda (changes requested):
This looks good. https://github.com/ModProg/derive-where/pull/33/files#r780831274 is the only thing really left to do, the rest are nits.

src/lib.rs
ModProg: sounds like a reasonable thing, I already wondered why PhantomData was not used anymore

src/test/mod.rs
ModProg: well the thing I documented in the FIXME, if we want to not use `input_without_derive_attributes`.

src/test/mod.rs
```diff
@@ -7,18 +7,41 @@ mod use_case;
...
+		generics,
+		item,
+	} = Input::from_input(span, &item)?;
+
+	Ok(derive_wheres
+		.iter()
+		.flat_map(|derive_where| iter::repeat(derive_where).zip(&derive_where.traits))
+		.map(|(derive_where, trait_)| generate_impl(derive_where, trait_, &item, generics))
+		.collect())
 }
```
ModProg, line 47: due to now havin the whole impementation in `derive_where` to be able to return the item on error, this is only used for testing

src/lib.rs
ModProg: should be working now, but I moved `derive_where_internal` to test as it is no longer used here.

That means we are not testing this with unit tests anymore. But with integration tests still.

ModProg (commented):
Should be fine now, hopefully

ModProg:
we should document the restriction conserning renaming, and maybe run a test how useful the error message in such a case will be.

ModProg:
And like I said test what happens if put the attribute_macro on a non struct/enum/unit.

.config/topic.dic
```diff
@@ -3,9 +3,11 @@ Changelog
 CHANGELOG
 destructure/G
 enum/S
+FIXME
 FQS
 io
 MSRV
+proc_macro_attribute
```
daxpedda, line 10: These additions can probably be removed again.

non-msrv-tests/tests/ui/union.stderr
```diff
@@ -1,21 +1,20 @@
...
   |
-6 | / #[derive_where(Debug)]
-7 | | union UnsupportedTrait<T> {
-8 | |     a: Option<PhantomData<T>>,
-9 | | }
-  | |_^
+5 | #[derive_where(Debug)]
+  | ^^^^^^^^^^^^^^^^^^^^^^
+  |
+  = note: this error originates in the attribute macro `derive_where` (in Nightly builds, run with -Z macro-backtrace for more info)
```
daxpedda, line 7: This seems to be wrong. I think the wrong span was passed down here.

non-msrv-tests/tests/ui/item.stderr
```diff
@@ -1,53 +1,51 @@
...
- --> tests/ui/item.rs:6:1
+error: empty `derive_where` found
+ --> tests/ui/item.rs:5:1
   |
-6 | struct NoAttribute<T>(PhantomData<T>);
-  | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
-
-error: unexpected option syntax
- --> tests/ui/item.rs:9:3
+5 | #[derive_where]
```
daxpedda, line 4: Again, this seems consistent that not the correct item span is passed down.

non-msrv-tests/tests/ui/default.stderr
```diff
@@ -1,38 +1,37 @@
...
 error: required `default` option on a variant if `Default` is being implemented
-  --> tests/ui/default.rs:17:1
+  --> tests/ui/default.rs:14:1
    |
-17 | / #[derive_where(Default)]
-18 | | enum MissingDefault<T> {
-19 | |     A(PhantomData<T>),
-20 | | }
-   | |_^
+14 | #[derive_where(Default)]
```
daxpedda, line 16: Here too.

non-msrv-tests/tests/ui/use_case.stderr
```diff
@@ -1,57 +1,61 @@
...
-  --> tests/ui/use_case.rs:16:1
+  --> tests/ui/use_case.rs:12:1
+   |
+12 | #[derive_where(Clone)]
+   | ^^^^^^^^^^^^^^^^^^^^^^
    |
-16 | / #[derive_where(Clone)]
-17 | | struct UnionEmpty {}
-   | |____________________^
+   = note: this error originates in the attribute macro `derive_where` (in Nightly builds, run with -Z macro-backtrace for more info)
```
daxpedda, line 31: All of these don't have the right span.

src/lib.rs
```diff
@@ -409,32 +399,53 @@ const DERIVE_WHERE: &str = "derive_where";
...
+		}
+	};
+	let cleaned_item = input_without_derive_where_attributes(item.clone());
 
-	Ok(derive_wheres
-		.iter()
-		.flat_map(|derive_where| iter::repeat(derive_where).zip(&derive_where.traits))
-		.map(|(derive_where, trait_)| generate_impl(derive_where, trait_, &item, generics))
-		.collect())
+	match { Input::from_input(span, &item) } {
```
daxpedda, line 428: I think the issue is actually here. This span references something that is probably dead, because we modified it? You can try taking the span from the cleaned item and see what happens.

src/test/mod.rs
daxpedda: This looks good to me, not a problem I believe.

daxpedda (changes requested):
The span issue is the only thing remaining.

> we should document the restriction conserning renaming, and maybe run a test how useful the error message in such a case will be.

> And like I said test what happens if put the attribute_macro on a non struct/enum/unit.

We could also do these in a separate PR, however you prefer.

non-msrv-tests/tests/ui/use_case.stderr
```diff
@@ -1,57 +1,53 @@
...
 
 error: derive-where doesn't support empty items, as this can already be handled by standard `#[derive(..)]`
-  --> tests/ui/use_case.rs:16:1
+  --> tests/ui/use_case.rs:13:1
    |
-16 | / #[derive_where(Clone)]
-17 | | struct UnionEmpty {}
-   | |____________________^
+13 | struct UnionEmpty {}
+   | ^^^^^^^^^^^^^^^^^^^^
```
ModProg, line 24: not identical to the original but shows the item now

src/lib.rs
ModProg: seams to have helped changing this!

non-msrv-tests/tests/ui/item.stderr
ModProg: But this is probably correct as this is about `derive_where` being empty

non-msrv-tests/tests/ui/use_case.stderr
daxpedda: I think that's fine.

daxpedda (approved):

daxpedda:
Feel free to merge or handle these last two points in this PR.

PR #36, opened 2022-01-21
Attribute proc-macro follow up

### LINKED ISSUE #35: Test and document usage errors/problems with attribute_macro
- [x] Renaming is only possible for the first macro, the rest are just attributes and setting their path does not work.
- [x] ~~Test error messages when attributes are renamed~~ (not applicable, because changing the attribute name actually works, but might behave unexpectedly, see #36)
- [x] Test error messages when applying on non derivable items

daxpedda:
Fixes #35.

I added a note in the README, but noticed that actually setting the path can produce some pretty unexpected behavior.

What do you think of adding an error if a `derive_where::derive_where` is found as not the first attribute? This isn't bulletproof of course because `derive_where_export::derive_where` could still remain undetected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,55 +34,12 @@ struct Example<T>(PhantomData<T>);
 ```
 
 Multiple `derive_where` attributes can be added to an item, but only the
-first one should use any path qualifications. Otherwise helper attributes
-won't be applied to any but the first `derive_where` attribute on the item.
+first one must use any path qualifications.
 
 ```rust
-#[derive_where::derive_where(Clone, Debug; T)]
-#[derive_where(PartialEq)]
-struct Example1<T, U> {
-	a: u8,
-	#[derive_where(skip)]
-	b: u8,
-	ph: PhantomData<(T, U)>,
-}
-
-let var1 = Example1 {
-	a: 42,
-	b: 42,
-	ph: PhantomData::<((), ())>,
-};
-let var2 = Example1 {
-	a: 42,
-	b: 0,
-	ph: PhantomData::<((), ())>,
-};
-
-// Field `b` is not compared, as expected.
-assert_eq!(var1, var2);
-
-#[derive_where::derive_where(Clone, Debug; T)]
-#[derive_where::derive_where(PartialEq)]
-struct Example2<T, U> {
-	a: u8,
-	#[derive_where(skip)]
-	b: u8,
-	ph: PhantomData<(T, U)>,
-}
-
-let var1 = Example2 {
-	a: 42,
-	b: 42,
-	ph: PhantomData::<((), ())>,
-};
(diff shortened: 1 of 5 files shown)


daxpedda:
I just stumbled on an idea. We could leave a marker to signify that this item was already handled by a derive-where macro. We control the outputted item, so we could just add a comment: `/* derive_where was here! */`. If that comment was found by a follow up proc macro attribute it could error out.

daxpedda:
Actually I couldn't find a way to mark the item. Comments aren't represented in the AST apparently, only doc comments, and I can't figure out how to make a "hidden" doc comment, so it doesn't show up in the documentation by the user. 

The other idea I had while writing this was to introduce another attribute called `derive_where_visited`, but it actually hits the same problem, it won't work without proper qualifying it unless it's imported by the user. We also don't have `$crate` for proc macros, see rust-lang/rust#54363, which would solve that problem.

So I think the best would be to introduce a hidden doc comment or find some other attribute that can be no-op. If you come up with anything hit me.

daxpedda:
To solve the problem with `derive_where_visited` we could also introduce a `crate` setting: `#[derive_where_::derive_where(crate = "derive_where_", Clone; T)]`?

daxpedda:
Let's figure out #37 before we decide to merge this.

ModProg (approved):
Looks good AFAICT. If attribute macros on fields were introduced, we could circumvent all of this, probably.

daxpedda:
Alright, re-based and adjusted to the latest changes we did in #37.

daxpedda:
Build failed because of some connection issue, would be nice if you re-start it @ModProg.

README.md
```diff
@@ -23,6 +23,25 @@ This will generate trait implementations for `Example` for any `T`,
...
+won't be applied to any but the first `derive_where` attribute on the item.
+
+```rust
+#[derive_where::derive_where(Clone, Debug)]
+#[derive_where(Default, Hash)]
+struct Example<T>(PhantomData<T>);
+```
+
+Multiple `derive_where` attributes can be added to an item, but only the
+first one must use any path qualifications.
```
ModProg, line 37: That sounds a bit repetitive, also half is missing in the docs

ModProg (changes requested):

README.md
daxpedda: This is wrong ... fixing it now.

daxpedda: Oh, I don't know where you got this from, but this is outdated?

daxpedda: No it is not ... I don't know whats happening ...

daxpedda: All right, I didn't see that it was duplicate ... it's good now, thanks!

ModProg (approved):

PR #37, opened 2022-01-21
Implement placeholder attribute to prevent multiple `derive_where`s with different paths

daxpedda:
This addresses a suggestion I made in #36. It implements a placeholder attribute that is applies when a `derive_where` attribute is processed. Any subsequent calls to `derive_where` with different paths will error out when they find this attribute.

Additionally:
- Adds a `#[derive_where(crate = "derive_where_")]` option.
- Fixes some typos.
- Adds a check for `#[derive_where(Zeroize(crate = "::zeroize"))]` because this is already the default.
- Update zeroize.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 ### Added
-- Support [`ZeroizeOnDrop`](https://docs.rs/zeroize/1.5.0-pre/zeroize/trait.ZeroizeOnDrop.html).
+- Support [`ZeroizeOnDrop`](https://docs.rs/zeroize/1.5.0/zeroize/trait.ZeroizeOnDrop.html).
 
 ### Removed
 - **Breaking Change**: Remove support for `Zeroize(drop)`.
--- a/non-msrv-tests/tests/ui/item.stderr
+++ b/non-msrv-tests/tests/ui/item.stderr
@@ -27,63 +27,69 @@ error: expected path, expected identifier
    |                        ^^^^^^^^^^^^^
 
 error: the `crate` option has to be defined in it's own `#[derive_where(..)` attribute
-  --> tests/ui/item.rs:17:16
+  --> tests/ui/item.rs:20:16
    |
-17 | #[derive_where(crate = "derive_where_", crate = "derive_where_")]
+20 | #[derive_where(crate = "derive_where_", crate = "derive_where_")]
    |                ^^^^^
 
-error: no traits found to implement, use `#[derive_where(..)` to specify some
-  --> tests/ui/item.rs:21:1
+error: duplicate `crate` option
+  --> tests/ui/item.rs:24:16
    |
-21 | struct OnlyCrate<T>(PhantomData<T>);
+24 | #[derive_where(crate = "derive_where_")]
+   |                ^^^^^^^^^^^^^^^^^^^^^^^
+
+error: no traits found to implement, use `#[derive_where(..)` to specify some
+  --> tests/ui/item.rs:28:1
+   |
+28 | struct OnlyCrate<T>(PhantomData<T>);
    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
 
 error: unnecessary path qualification, `::derive_where` is used by default
-  --> tests/ui/item.rs:23:24
+  --> tests/ui/item.rs:30:24
    |
-23 | #[derive_where(crate = "::derive_where")]
+30 | #[derive_where(crate = "::derive_where")]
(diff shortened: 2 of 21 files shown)


CHANGELOG.md
```diff
@@ -6,7 +6,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
```
ModProg, line 7: We should also document the whole change to procmacro :D

non-msrv-tests/tests/ui/item.stderr
```diff
@@ -14,38 +14,68 @@ error: empty `derive_where` found
   |
   = note: this error originates in the attribute macro `derive_where` (in Nightly builds, run with -Z macro-backtrace for more info)
 
+error: the `crate` option has to be defined in it's own `#[derive_where(..)` attribute
+  --> tests/ui/item.rs:11:16
+   |
+11 | #[derive_where(crate(derive_where_))]
+   |                ^^^^^
```
ModProg, line 21: This looks a bit strange, shouldn't the error be:

> crate has do be defined as `crate="smth"` and not `crate(smth)`?

Because it is in its own derive_where attribute

src/error.rs
```diff
@@ -6,6 +6,34 @@ use proc_macro2::Span;
...
 
 impl Error {
+	/// `derive_where` was already applied on this item before.
+	pub fn visited(span: Span) -> syn::Error {
+		syn::Error::new(
+			span,
+			"`#[derive_where(..)` was already applied to this item before, this occurs when using \
+			 a qualified path for any `#[derive_where(..)`s except the first",
+		)
+	}
```
ModProg, line 16: Is there a UI test for this? or did I just overlook it?

non-msrv-tests/tests/ui/item.stderr
```diff
@@ -14,38 +14,68 @@ error: empty `derive_where` found
...
+   |
+14 | #[derive_where(crate = "struct Test")]
+   |                        ^^^^^^^^^^^^^
+
+error: the `crate` option has to be defined in it's own `#[derive_where(..)` attribute
+  --> tests/ui/item.rs:17:16
+   |
+17 | #[derive_where(crate = "derive_where_", crate = "derive_where_")]
+   |                ^^^^^
+
```
ModProg, line 34: I was thinking if this should be a seperate error `duplicate_crate`, but it's a pretty stupid misstake to make, and I'd guess you'd see the mistake if it is marked as a mistake. Just maybe the span should be the whole content and not only the first `crate`. 

Not sure tho.

ModProg (changes requested):

CHANGELOG.md
daxpedda: I planned to leave that to the other PR: #36.

non-msrv-tests/tests/ui/item.stderr
daxpedda: Yes, I forgot to mention, basically I was sick of creating new code branches to handle errors. The code is becoming really bloated because of it. I will look into it again.

src/error.rs
daxpedda: No, I somehow missed that.

non-msrv-tests/tests/ui/item.stderr
daxpedda: As above, awful error handling code branches.

daxpedda: Yeah, I tried, but I couldn't really figure out a clean way without changing a lot of stuff around. Unless you really mind I plan to leave it like that.

ModProg: No I think we can keep it as is.

ModProg: Maybe we can document it in the test case, as not ideal but the best we can easily do here.

src/error.rs
ModProg: did you add one? because this was resolved but I could not find it?

ModProg (changes requested):

non-msrv-tests/tests/ui/item.stderr
daxpedda: That sounds like a good idea.

src/error.rs
daxpedda: https://github.com/ModProg/derive-where/pull/37/files#diff-b4e400fe186fc9514f7d7023c38062fcb10c451c6f5111a7f241f646e1d8f390R83-R89
here

ModProg: oh ok, nvm

non-msrv-tests/tests/ui/item.stderr
daxpedda: Done.

ModProg (approved):

PR #38, opened 2022-01-21
Improve code generation of `Ord` and `PartialOrd`

daxpedda:
See code generation tests, should be very self-explanatory.

### DIFF
--- a/src/trait_/common_ord.rs
+++ b/src/trait_/common_ord.rs
@@ -105,35 +105,48 @@ pub fn build_ord_signature(item: &Item, trait_: &DeriveTrait, body: &TokenStream
 				// other. The index for these variants is used to determine which
 				// `Ordering` to return.
 				for (index, variant) in variants.iter().enumerate() {
-					let mut arms = Vec::with_capacity(variants.len() - 1);
-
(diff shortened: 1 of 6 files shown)


src/trait_/common_ord.rs
```diff
@@ -105,35 +105,48 @@ pub fn build_ord_signature(item: &Item, trait_: &DeriveTrait, body: &TokenStream
...
-						// Make sure we aren't comparing the same variant with itself.
-						if index != index_other {
-							use std::cmp::Ordering::*;
+					// The first variant is always `Less` then everything.
+					if index == 0 {
+						different.push(quote! {
+							#pattern => #less,
+						})
+					}
+					// The last variant is always `Less` then everything.
```
ModProg, line 116:
```suggestion
					// The last variant is always `Greater` then everything.
```

ModProg (changes requested):
looks good :+1:

src/trait_/common_ord.rs
daxpedda: Done.

ModProg (approved):

PR #39, opened 2022-01-23
Publish v1.0.0-rc.2

daxpedda:
Builds on #36.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -14,7 +14,7 @@ license = "MIT OR Apache-2.0"
 name = "derive-where"
 readme = "README.md"
 repository = "https://github.com/ModProg/derive-where"
-rust-version = "1.34.2"
+rust-version = "1.34"
 version = "1.0.0-rc.2"
 
 [lib]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -14,7 +14,7 @@ license = "MIT OR Apache-2.0"
 name = "derive-where"
 readme = "README.md"
 repository = "https://github.com/ModProg/derive-where"
-rust-version = "1.34.2"
+rust-version = "1.34"
```
daxpedda, line 17: I noticed that skipping the minor version seems to very common in other crates right now, it looks reasonable to me and is more correct anyway. 1.34.2 doesn't actually fix anything that concerns us.

daxpedda:
Another connection issue in the CI.

Cargo.toml
ModProg: sounds good

daxpedda:
Re-based and good to go @ModProg.

ModProg (approved):

ModProg:
only the date is in the past :smile:

PR #40, opened 2022-02-19
Derive forwarding

daxpedda:
This forwards the attribute proc-macro to a derive proc-macro, enabling correct `cfg` evaluation.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -360,11 +360,11 @@ mod util;
 
 use std::{borrow::Cow, iter};
 
-use proc_macro2::{Ident, Span, TokenStream};
-use quote::{quote, ToTokens};
+use proc_macro2::TokenStream;
+use quote::{quote, quote_spanned, ToTokens};
 use syn::{
 	spanned::Spanned, Attribute, DataEnum, DataStruct, DataUnion, DeriveInput, Fields, FieldsNamed,
-	FieldsUnnamed, Generics, Variant,
+	FieldsUnnamed, Generics, Lit, Meta, NestedMeta, Path, Variant,
 };
 
 #[cfg(feature = "zeroize")]
@@ -379,8 +379,10 @@ use self::{
 	util::Either,
 };
 
-/// Name of the `derive_where` proc-macro.
+/// Name of the `derive_where` attribute proc-macro.
 const DERIVE_WHERE: &str = "derive_where";
+/// Name of the `DeriveWhere` derive proc-macro.
+const DERIVE_WHERE_INTERNAL: &str = "DeriveWhere";
 /// Name of the `derive_where_visited` proc-macro.
 const DERIVE_WHERE_VISITED: &str = "derive_where_visited";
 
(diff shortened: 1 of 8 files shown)


src/lib.rs
```diff
@@ -412,61 +427,149 @@ const DERIVE_WHERE_VISITED: &str = "derive_where_visited";
...
+			}
+		},
+		Err(error) => {
+			original_input.extend(error.into_compile_error());
+			original_input.into()
+		}
+	}
+}
+
+/// Convenient way to deal with [`Result`] for [`derive_where`].
```
ModProg, line 456: Apparently this `Result` needs to be explicitly linked

tests/cfg.rs
```diff
@@ -0,0 +1,13 @@
+use std::marker::PhantomData;
+
+use derive_where::derive_where;
+
+#[test]
+fn cfg() {
+	#[derive_where(Clone)]
+	struct Test<T> {
+		a: PhantomData<T>,
+		#[cfg(invalid)]
```
ModProg, line 10: Seeing this got me thinking. We could also copy all cfg attributes throughout our expansion. But that would increase complexity a lot probably and would make it necessary to adapt every expansion probably.

ModProg (changes requested):
Just the doc comment i guess, rest looks fine.

src/lib.rs

tests/cfg.rs
daxpedda: Agreed, I hope this solution works.

daxpedda:
I still have to figure out MSRV. Worst case scenario we upgrade the MSRV to 1.52.

ModProg (changes requested):
Probably just need to resort that one macro invocation for the test to succeed

ModProg:
> I still have to figure out MSRV. Worst case scenario we upgrade the MSRV to 1.52.

Oh right, I remember. Well below 1.52 then just use the derive macro diectly I guess.

daxpedda:
Currently the issue is `derive_where_visited`, it can't be re-ordered because it would make it simply ineffective. So unless I find a different solution:
- We remove `derive_where_visited` and hope that users don't do anything stupid.
- We bump MSRV.

I'm currently leaning towards bumping MSRV, WDYT?

daxpedda:
Apparently `derive`s can't actually produce attribute proc-macros until version 1.57 ...

ModProg:
> I'm currently leaning towards bumping MSRV, WDYT?

I mean, we have a version that works with the lower RV so if someone needs that, they can stay on that version.

> - We remove derive_where_visited and hope that users don't do anything stupid.

I think we can keep this in mind and if needed add a `low-msrv` feature flag that removes the helpfull errors but allows to run on a lower version.

daxpedda:
Sry, didn't see your comment before I pushed. I just did a full upgrade, can revert it of course. Also fixed some zeroize links.



> I mean, we have a version that works with the lower RV so if someone needs that, they can stay on that version.

I don't think it does work, not evaluating `cfg` is simply broken imo:
```rust
#[derive_where(Clone)]
struct Test<T> {
	a: PhantomData<T>,
	#[cfg(test)]
	b: u8,
}
```
The fact that this can't compile, seems pretty broken to me.

> I think we can keep this in mind and if needed add a `low-msrv` feature flag that removes the helpfull errors but allows to run on a lower version.

I don't mind, but I think we should wait until somebody actually needs this. Personally I needed 1.51, but I don't need it anymore. Am happy to implement it of course if you think it's important.

ModProg:
> I don't mind, but I think we should wait until somebody actually needs this. Personally I needed 1.51, but I don't need it anymore. Am happy to implement it of course if you think it's important.

I would just add a comment about this in the documentation maybe, but I agree that it makes little sense implementing a feature no one currently has a need for.

ModProg:
Did this just break inside the std? 
https://github.com/ModProg/derive-where/runs/5264501391?check_suite_focus=true#step:6:191

daxpedda:
> Did this just break inside the std? https://github.com/ModProg/derive-where/runs/5264501391?check_suite_focus=true#step:6:191

Yeah, already fixed that, the issue is that Miri doesn't have support for a lot of things inside of `std`, in this case it was `trybuild`, so I disabled UI tests for Miri, which isn't what we were trying to test with Miri anyway.

daxpedda:
> I would just add a comment about this in the documentation maybe, but I agree that it makes little sense implementing a feature no one currently has a need for.

Done. I'm gonna open an issue where interested parties can note that they need a lower MSRV.

ModProg (approved):
looks good 👍

PR #44, opened 2022-02-22
Enforce `Eq` trait on field types

### LINKED ISSUE #43: Support `skip` for `Eq`

daxpedda:
Fixes #43. It does not only allow `skip`ing, but actually enforces `Eq` on every type like `std` does.

### DIFF
--- a/src/trait_/eq.rs
+++ b/src/trait_/eq.rs
@@ -1,6 +1,9 @@
 //! [`Eq`](std::cmp::Eq) implementation.
 
-use crate::{DeriveTrait, TraitImpl};
+use proc_macro2::TokenStream;
+use quote::quote;
+
+use crate::{Data, DeriveTrait, Item, TraitImpl};
 
 /// Dummy-struct implement [`Trait`](crate::Trait) for [`Eq`](std::cmp::Eq).
 pub struct Eq;
@@ -12,5 +15,39 @@ impl TraitImpl for Eq {
 
 	fn default_derive_trait(&self) -> DeriveTrait {
 		DeriveTrait::Eq
+	}
+
+	fn supports_skip(&self) -> bool {
+		true
+	}
+
+	fn build_signature(
+		&self,
+		_item: &Item,
+		_traits: &[DeriveTrait],
+		_trait_: &DeriveTrait,
+		body: &TokenStream,
+	) -> TokenStream {
+		quote! {
+			#[inline]
+			fn assert_receiver_is_total_eq(&self) {
+				struct __AssertEq<__T: ::core::cmp::Eq + ?::core::marker::Sized>(::core::marker::PhantomData<__T>);
+
+				#body
+			}
+		}
+	}
+
+	fn build_body(
+		&self,
+		_traits: &[DeriveTrait],
+		trait_: &DeriveTrait,
+		data: &Data,
+	) -> TokenStream {
+		let types = data.iter_fields(trait_).map(|field| field.type_);
+
+		quote! {
+			#(let _: __AssertEq<#types>;)*
+		}
 	}
 }
--- a/src/data/field.rs
+++ b/src/data/field.rs
@@ -4,7 +4,7 @@ use std::fmt::{self, Display, Formatter};
 
 use proc_macro2::{Span, TokenStream};
 use quote::{format_ident, IdentFragment, ToTokens, TokenStreamExt};
-use syn::{Attribute, FieldsNamed, FieldsUnnamed, Ident, Index, Result};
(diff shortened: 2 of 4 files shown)


ModProg:
> I believe it makes sense to automatically skip whatever `PartialEq` skips, what do you think?

yeah, maybe we should generally group things like `Eq`, `PartialEq`, `ParialOrd`, `Ord` and `Hash`  as they should always ignore the same right? 

Not entirely sure related to the ordering but generally `Eq` means `Ord` should be Equal and `Hash` should be the same as well.

ModProg:
Maybe there are other Traits where this also makes sense?

src/trait_/eq.rs
```diff
@@ -13,4 +16,38 @@ impl TraitImpl for Eq {
...
+	fn build_signature(
+		&self,
+		_item: &Item,
+		_traits: &[DeriveTrait],
+		_trait_: &DeriveTrait,
+		body: &TokenStream,
+	) -> TokenStream {
+		quote! {
+			#[inline]
+			fn assert_receiver_is_total_eq(&self) {
```
ModProg, line 33: do we need to add the doc hidden as well?

src/trait_/eq.rs
```diff
@@ -13,4 +16,38 @@ impl TraitImpl for Eq {
...
+				struct __AssertEq<__T: ::core::cmp::Eq + ?::core::marker::Sized>(::core::marker::PhantomData<__T>);
```
ModProg, line 34: technically we could also have our own global `__AssertEq` lying around and use it, but that would then depend on being able to know our path.

ModProg:
> The reason why I was proposing `Eq` to take in account `skip`s for `PartialEq` is because `Eq` doesn't have it's own implementation, it uses the implementation of `PartialEq`. Basically `Eq` is just a marker trait for `PartialEq`.
> 
> I don't think we should group any others together because they aren't really the same, but I understand where you are coming from, this isn't an uncommon use-case. But I think `skip` without specifying any specific trait already covers that. Otherwise we could introduce grouped `skip`s somehow.

yeah, but IIRC Ord, PartialOrd and Hash have requirements on how Eq has to behave, so generally I think it would make sense to have some way of saying skip for all comparisons. 

The usefulness really depends on how often you want to skip all of those, but not skip them for other traits.

src/trait_/eq.rs
daxpedda: You can't export types from a proc-macro crate sadly, the only way we could do that is if we make another non-proc-macro crate that holds these types and re-exports the proc-macro from another crate. Basically we would have two crates: `derive-where` and `derive-where-macro`. Should we do that?

src/trait_/eq.rs
daxpedda: Good question, I will check.

daxpedda:
> yeah, but IIRC Ord, PartialOrd and Hash have requirements on how Eq has to behave, so generally I think it would make sense to have some way of saying skip for all comparisons.

`Hash` doesn't require any of the others. In any case, there is already way of skipping all, that's just using `#[derive_where(skip)]`, this skips `Eq`, `Hash`, `Ord`, `PartialEq`, `PartialOrd`, `Zeroize` and `ZeroizeOnDrop`. The only trait that includes that might be undesirable is `Zeroize` and `ZeroizeOnDrop`. We could introduce a group, like `#[derive_where(skip(AllButZeroize))]` (requires some serious bikesheding).

src/trait_/eq.rs
daxpedda: It doesn't seem to be necessary, doesn't show up in docs at all.

ModProg:
> `Hash` doesn't require any of the others.

What I meant was not `Hash` depends on `Eq` but if `Eq` is also implemented, it needs to behave the same way: https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq

ModProg:
But I agree, that there is probably no need to support skipping all but one or two in 90% of cases, at least with the current limited set of supported traits

src/trait_/eq.rs
ModProg: oh right. If this would be the only reason to do so, I would say no. If we were to add a wrapper crate for some reason, we can do this.

ModProg (approved):

daxpedda:
> > `Hash` doesn't require any of the others.
> 
> What I meant was not `Hash` depends on `Eq` but if `Eq` is also implemented, it needs to behave the same way: https://doc.rust-lang.org/std/hash/trait.Hash.html#hash-and-eq

Ah, I think now I see what you were talking about until now ... this is a really good point. I'm gonna merge this for now and we should discuss this separately.

PR #47, opened 2022-03-07
Skip groups

daxpedda:
As discussed on Discord. I also made `Trait` `Copy` on the way, not sure why I didn't before.


ModProg:
Nice, I'll look into it tomorrow!

PR #50, opened 2022-08-09
Improve Examples

daxpedda:
I discovered in https://github.com/lu-zero/derive_bounded/issues/2 that the examples on the front page might lead users to believe you have to write a `#[derive_where(..)]` for each trait, I improved the examples to not only use one trait.

Also fixed some wording about `Default` on `enum`s, Rusts supports this now.

EDIT: Looking at the README again, I noticed a lot of links were broken and missing, so I fixed that too.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,12 +6,12 @@
 
 ## Description
 
-Derive macro to simplify deriving standard and other traits with custom
-generic type bounds.
+Attribute proc-macro to simplify deriving standard and other traits with
+custom generic type bounds.
 
 ## Usage
 
-The `derive_where` macro can be used just like std's `#[derive(...)]`
+The `derive_where` attribute can be used just like std's `#[derive(...)]`
 statements:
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -6,12 +6,12 @@
 
 ## Description
 
-Derive macro to simplify deriving standard and other traits with custom
-generic type bounds.
+Attribute proc-macro to simplify deriving standard and other traits with
+custom generic type bounds.
```
daxpedda, line 10: Since we moved to using an attribute macro instead of `derive`, we didn't update the wording. @ModProg I think it would be good if we also updated the description of the GitHub repo with the same text.

ModProg: good catch

ModProg:
I changed the Description, is this ready to merge than?

daxpedda:
Yeah, feel free to merge anytime!

PR #54, opened 2022-12-25
Make CI green again

ModProg:


src/lib.rs
```diff
@@ -2,7 +2,12 @@
 #![cfg_attr(feature = "nightly", feature(allow_internal_unstable))]
 #![allow(clippy::tabs_in_doc_comments)]
 #![warn(clippy::cargo, clippy::missing_docs_in_private_items)]
-#![cfg_attr(doc, warn(rustdoc::all), allow(rustdoc::missing_doc_code_examples))]
+#![cfg_attr(
+	all(doc, feature = "nightly"),
+	warn(rustdoc::all),
+	allow(rustdoc::missing_doc_code_examples),
+	feature(rustdoc_missing_doc_code_examples)
+)]
```
daxpedda, line 10:
```suggestion
#![cfg_attr(doc, allow(unknown_lints), warn(rustdoc::all))]
```

I think this is more appropriate, we want to keep testing doc lints in stable too after all. It's kinda unfortunate that `rustdoc::all` includes an unstable lint that triggers `unknown_lints` even if you are on stable.

daxpedda (changes requested):
Otherwise LGTM.

src/lib.rs
ModProg: Makes sense. Maybe something they could change `all`, doesn't make sense to include an unavailable lint IMO.

daxpedda (approved):

PR #60, opened 2023-03-18
Update to `syn` v2

daxpedda:
The big change here is that we have to parse `MetaList` on our own.

One thing I was concerned about was that the new `NestedMeta` used to support a free standing literal, but when you parse to `Meta` that's an error now. Luckily we didn't use it anywhere.

Some slightly improved error messages.

One thing I noticed we can improve now is that `MetaNameValue` now takes an `Expr` as a value, so now we can actually take a path instead of a string literal in the `crate` attribute.

Also noticed that our spellchecker wasn't running properly but only checking `CHANGELOG.md`.

### DIFF
--- a/src/attr/item.rs
+++ b/src/attr/item.rs
@@ -7,7 +7,7 @@ use syn::{
 	parse::{discouraged::Speculative, Parse, ParseStream},
 	punctuated::Punctuated,
 	spanned::Spanned,
-	Attribute, Data, Ident, Meta, NestedMeta, Path, PredicateType, Result, Token, TraitBound,
+	Attribute, Data, Ident, Meta, Path, PredicateType, Result, Token, TraitBound,
 	TraitBoundModifier, Type, TypeParamBound, TypePath, WhereClause, WherePredicate,
 };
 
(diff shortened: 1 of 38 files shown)


src/lib.rs
```diff
@@ -481,43 +481,45 @@ fn derive_where_internal(mut item: DeriveInput) -> Result<TokenStream, syn::Erro
...
-												}
-												None => crate_ = Some(path),
-											}
-										}
+								let path = match &name_value.value {
+									Expr::Lit(ExprLit {
+										lit: Lit::Str(lit_str),
+										..
+									}) => match lit_str.parse::<Path>() {
+										Ok(path) => path,
```
ModProg, line 499: We could do a warning, that it is deprecated on nightly IIRC, but probably not worth it

src/trait_/zeroize_on_drop.rs
```diff
@@ -18,38 +21,40 @@ impl TraitImpl for ZeroizeOnDrop {
...
-							} else {
-								return Err(Error::option_syntax(name_value.lit.span()));
+								},
+								Expr::Path(ExprPath { path, .. }) => path.clone(),
+								_ => return Err(Error::option_syntax(name_value.value.span())),
+							};
+
+							if path == util::path_from_strs(&["zeroize"]) {
+								return Err(Error::path_unnecessary(path.span(), "::zeroize"));
 							}
```
ModProg, line 55: Would it make sense to reuse code between this and the `derive_where` `crate`?

tests/ui/not-zeroize/item_option_syntax.stderr
```diff
@@ -0,0 +1,17 @@
...
+ --> tests/ui/not-zeroize/item_option_syntax.rs:5:18
+  |
+5 | #[derive_where = "invalid"]
+  |                  ^^^^^^^^^
+
+error[E0601]: `main` function not found in crate `$CRATE`
+ --> tests/ui/not-zeroize/item_option_syntax.rs:6:48
+  |
+6 | struct WrongAttributeSyntax<T>(PhantomData<T>);
+  |                                                ^ consider adding a `main` function to `$DIR/tests/ui/not-zeroize/item_option_syntax.rs`
```
ModProg, line 17: we could add a `main` function

tests/ui/zeroize/item_option_syntax.stderr
```diff
@@ -0,0 +1,17 @@
...
+ --> tests/ui/zeroize/item_option_syntax.rs:5:18
+  |
+5 | #[derive_where = "invalid"]
+  |                  ^^^^^^^^^
+
+error[E0601]: `main` function not found in crate `$CRATE`
+ --> tests/ui/zeroize/item_option_syntax.rs:6:48
+  |
+6 | struct WrongAttributeSyntax<T>(PhantomData<T>);
+  |                                                ^ consider adding a `main` function to `$DIR/tests/ui/zeroize/item_option_syntax.rs`
```
ModProg, line 17: ^^

ModProg (changes requested):

src/attr/zeroize_fqs.rs
```diff
@@ -26,24 +26,26 @@ impl ZeroizeFqs {
...
-									self.0 = true
-								}
+				for meta in &nested {
+					if let Meta::Path(path) = meta {
+						if path.is_ident(Self::FQS) {
+							if self.0 {
+								return Err(Error::option_duplicate(path.span(), Self::FQS));
 							} else {
-								return Err(Error::option(path.span()));
+								self.0 = true
```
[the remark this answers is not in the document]
daxpedda: No, but it definitely looks that way, the nesting has just moved a bit.
I was a bit shocked myself^^.

src/lib.rs
daxpedda: Last time I checked it's still not possible to insert warnings inside a proc-macro, so it would have to be an error.
I'm against an error, but a warning would be nice even on non-nightly.

tests/ui/not-zeroize/item_option_syntax.stderr
daxpedda: Ah, lol.

src/trait_/zeroize_on_drop.rs
daxpedda: That's a good idea, I will give it a shot.

src/attr/skip.rs
```diff
@@ -78,8 +78,11 @@ impl Skip {
...
 			}
 			Meta::List(list) => {
+				let nested =
+					list.parse_args_with(Punctuated::<Meta, Token![,]>::parse_terminated)?;
+
 				// Don't allow an empty list.
-				if list.nested.is_empty() {
+				if nested.is_empty() {
 					return Err(Error::option_empty(list.span()));
 				}
```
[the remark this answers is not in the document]
daxpedda: Yes, I agree.

src/attr/item.rs
```diff
@@ -33,61 +33,57 @@ impl ItemAttr {
...
+
+									// Don't parse `Skip` yet, because it needs access to all
+									// `DeriveWhere`s.
+									skip_inners.push(meta);
+								} else if meta.path().is_ident(Incomparable::INCOMPARABLE) {
+									// Needs to be parsed after all traits are known.
+									incomparables.push(meta)
+								} else if meta.path().is_ident("crate") {
+									// Do nothing, we checked this before
+									// already.
```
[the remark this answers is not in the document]
ModProg: isn't the comments a few lines up even wider ... strange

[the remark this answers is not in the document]
daxpedda: Yeah it's weird.

src/lib.rs
ModProg: Well there is on nightly: https://doc.rust-lang.org/stable/proc_macro/struct.Diagnostic.html#method.span_warning, but not sure how well that works.

src/trait_/zeroize_on_drop.rs
daxpedda: After a pathetic attempt I didn't really like it. Feel free to give it a shot yourself.

src/lib.rs
daxpedda: Ah, right. Never tried it too. In any case, I'm fine with not adding a nightly-only warning.

src/attr/item.rs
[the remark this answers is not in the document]
ModProg: Well it seems to behave differently when it is just a comment and nothing else, i.e.
```rs
									// Do nothing, we checked this before already.
									()

```
would not break.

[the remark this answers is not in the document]
daxpedda: Ah, I see. Any preference? Personally I'm fine with how it is now, but happy to adapt whatever you prefer.

src/trait_/zeroize_on_drop.rs
ModProg: I'd be fine with keeping is as is then.

ModProg (approved):

PR #61, opened 2023-03-19
Add non-exhaustive `Debug` output for structs

### LINKED ISSUE #59: `Debug` and skip could use `finish_non_exhaustive()`
I noticed when implementing some `Debug`s by hand that when I want to use skip I also use [`finish_non_exhaustive()`](https://doc.rust-lang.org/nightly/std/fmt/struct.DebugStruct.html#method.finish_non_exhaustive).

What do you think?

daxpedda:
Somehow we only did integration tests on tuple structs, so I added some more to specifically test this feature.

On the way I learned a cool thing how you can skip tests without having to adjust your `Cargo.toml`, just hide them behind a folder, which I did now. This also fixes the Rustfmt issue.

~~Additionally I renamed the skip tests called `variant` to `enum`, kinda made more sense to me.~~
EDIT: Actually reverted this again, I think we did it because you can't actually use `skip_inner` on enums, only on the variants.

Fixes #59.


ModProg (changes requested):
We could have a test for struct enum variants.

daxpedda:
> We could have a test for struct enum variants.

Wow, completely forgot about that.

ModProg (approved):

PR #69, opened 2023-08-22
Prefix field names to avoid collisions

### LINKED ISSUE #68: derive_where(Debug) compile error if there's a struct parameter named `f`
```rust
use std::marker::PhantomData;
use derive_where::derive_where;

#[derive_where(Debug)]
pub struct Problematic<T> {
    f: String,
    _marker: PhantomData<T>,
}
```

results in
```txt
error[E0308]: mismatched types
    --> src/lib.rs:5:1
     |
5    | #[derive_where(Debug)]
     | ^^^^^^^^^^^^^^^^^^^^^^
     | |
     | types differ in mutability
     | arguments to this function are incorrect
     |
     = note: expected mutable reference `&mut Formatter<'_>`
                        found reference `&String`

```

The reason is that the expanded macro is trying to use a symbol `__f` in two conflicting ways: once for the struct member `f`, and once for the second argument to `Debug::fmt`. `cargo expand` shows that the example above expands to:

```rust
impl<T> ::core::fmt::Debug for Problematic<T> {
    fn fmt(&self, __f: &mut ::core::fmt::Formatter<'_>) -> ::core::fmt::Result {
        match self {
            Problematic { f: ref __f, _marker: ref ___marker } => {
                let mut __builder = ::core::fmt::Formatter::debug_struct(
                    __f,
                    "Problematic",
                );
                ::core::fmt::DebugStruct::field(&mut __builder, "f", __f);
                ::core::fmt::DebugStruct::field(&mut __builder, "_marker", ___marker);
                ::core::fmt::DebugStruct::finish(&mut __builder)
            }
        }
    }
}
```

The first argument to `debug_struct` is our `String`, `f`, where it was meant to be the `&mut Formatter`.

daxpedda:
Fixes #68.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,9 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+- Avoid collisions between field names and trait method parameters.
+
 ### Changed
 - `PartialOrd` implementations now use `Ord` if applicable.
 
--- a/src/data/field.rs
+++ b/src/data/field.rs
@@ -110,8 +110,8 @@ impl<'a> Field<'a> {
 		type_: &'a Type,
 	) -> Result<Self> {
 		let attr = FieldAttr::from_attrs(derive_wheres, skip_inner, attrs)?;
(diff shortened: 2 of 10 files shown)


daxpedda:
Blocked on #70.

PR #72, opened 2023-08-23
Fix `PartialOrd` implementations when using bounds

daxpedda:
See https://github.com/facebook/voprf/issues/119.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- Don't use `Ord` in `PartialOrd` implementations if using any bounds.
+
 ## [1.2.2] - 2023-08-22
 
 ### Fixed
--- a/src/trait_.rs
+++ b/src/trait_.rs
@@ -133,17 +133,25 @@ impl TraitImpl for Trait {
 
 	fn build_signature(
 		&self,
+		any_bound: bool,
 		item: &Item,
 		traits: &[DeriveTrait],
 		trait_: &DeriveTrait,
 		body: &TokenStream,
 	) -> TokenStream {
 		self.implementation()
-			.build_signature(item, traits, trait_, body)
+			.build_signature(any_bound, item, traits, trait_, body)
 	}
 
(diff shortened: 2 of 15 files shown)


src/trait_.rs
```diff
@@ -201,6 +210,7 @@ pub trait TraitImpl {
 	/// Build method body for this [`Trait`].
 	fn build_body(
 		&self,
+		_any_bound: bool,
```
daxpedda, line 213: Could have also passed `&[Generics]` directly, but I guess we can change that when we actually need it.

ModProg: yeah, I think this should be fine for now.

daxpedda:
I pushed the next release in here as well.

CHANGELOG.md
```diff
@@ -5,7 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
-## [1.2.2] - 2023-08-22
+## [1.2.3] - 2023-08-23
+
+### Fixed
+- Don't use `Ord` in `PartialOrd` implementations if using any bounds.
+
+## [1.2.2] - 2023-08-22 [YANKED]
```
ModProg, line 13: At first I thought this was supposed to be a link, and was confused where the address would be

src/trait_/partial_ord.rs
```diff
@@ -42,10 +43,16 @@ impl TraitImpl for PartialOrd {
...
+		trait_: &DeriveTrait,
+		data: &Data,
+	) -> TokenStream {
 		if data.is_empty(**trait_)
 			|| data.is_incomparable()
-			|| traits.iter().any(|trait_| trait_ == Trait::Ord)
+			|| (!any_bound && traits.iter().any(|trait_| trait_ == Trait::Ord))
 		{
 			TokenStream::new()
 		} else {
```
ModProg, line 58: Why do we even hit this branch, shouldn't it be `unreachable`?

CHANGELOG.md
daxpedda: Yeah, just following what https://keepachangelog.com/en/1.0.0/#yanked says.

src/trait_/partial_ord.rs
daxpedda: Ah, you mean why is the whole function being called in the first place?
That's because the caller has no awareness of whats going on, so the function is always called and the implementation has to figure out if it was necessary in the first place or not.

PR #76, opened 2023-08-31
Take representation into account for `(Partial)Ord`

daxpedda:
Would like to add more tests, but that's it.
On that note, I also improved a the `(Partial)Ord` generation on `safe`!


src/item.rs
```diff
@@ -88,3 +95,146 @@ impl Item<'_> {
...
+					return Err(Error::repr(attr.span()));
+				}
+			}
+		}
+
+		let is_unit = variants.iter().all(|variant| match &variant.fields {
+			Fields::Named(fields) => fields.named.is_empty(),
+			Fields::Unnamed(fields) => fields.unnamed.is_empty(),
+			Fields::Unit => true,
+		});
```
ModProg, line 145: Why not use https://docs.rs/syn/latest/syn/enum.Fields.html#method.is_empty

src/trait_/common_ord.rs
```diff
@@ -87,16 +94,16 @@ pub fn build_ord_signature(item: &Item, trait_: &DeriveTrait, body: &TokenStream
...
 
+				let path = trait_.path();
+				let method = match trait_ {
+					DeriveTrait::PartialOrd => quote! { partial_cmp },
+					DeriveTrait::Ord => quote! { cmp },
+					_ => unreachable!("unsupported trait in `prepare_ord`"),
+				};
+
 				// Nightly or unsafe (default) implementation.
 				#[cfg(any(feature = "nightly", not(feature = "safe")))]
```
ModProg, line 105: does this nesting still make sense?

src/trait_/common_ord.rs
```diff
@@ -106,98 +113,77 @@ pub fn build_ord_signature(item: &Item, trait_: &DeriveTrait, body: &TokenStream
...
+						},
+						Discriminant::UnitRepr(repr) => quote! {
+							#path::#method(
+								self as #repr,
+								__other as #repr,
+							)
+						},
+						Discriminant::Default | Discriminant::Repr(_) => {
+							build_recursive_order(trait_, variants, &incomparable)
 						}
```
ModProg, line 186: the only difference between safe and unsafe seems to be the handling of `Repr`, maybe it would make more sense to merge the two and only differentiate in that branch

tests/nightly.rs
ModProg, line 1: do we have tests for the unsafe on stable for the supported reprs?

ModProg (changes requested):

tests/nightly.rs
daxpedda: No, that's still missing. But FWIW it's stable stuff, so we don't have to worry about it breaking at some point.

src/item.rs
daxpedda: Done.

src/trait_/common_ord.rs
daxpedda: Removed while merging the other blocks like you said below.

src/trait_/common_ord.rs
daxpedda: Done.

ModProg (approved):

PR #77, opened 2023-09-01
Build manual discriminant order if discriminant isn't available

### LINKED ISSUE #73: Improve `nightly` crate feature documentation.
I just realized we didn't actually specify in the documentation that the `nightly` feature requires the nightly Rust version and is therefor unstable.

daxpedda:
Still have to fix and add more tests.
Documentation should also be adjusted.

Fixes #73.


src/lib.rs
```diff
@@ -623,18 +624,22 @@ fn generate_impl(
...
 	item: &Item,
-	generics: &Generics,
+	generics: &SplitGenerics,
 ) -> TokenStream {
-	let (impl_generics, type_generics, where_clause) = generics.split_for_impl();
-	let mut where_clause = where_clause.map(Cow::Borrowed);
+	let SplitGenerics {
+		r#impl,
+		ty,
+		r#where,
```
ModProg, line 632: I'd probably have used something like `imp` and `wher` or `impl_gens` and `where_clause` instead of the `r#`.

src/trait_/common_ord.rs
```diff
@@ -214,79 +231,131 @@ pub fn build_ord_signature(
...
+	for variant in variants {
+		let discriminant = if let Some(discriminant) = variant.discriminant {
+			if !has_non_isize
+				&& !matches!(
+					discriminant,
+					Expr::Lit(ExprLit {
+						lit: Lit::Int(_),
+						..
+					})
+				) {
```
ModProg, line 259: that does `-10` parse to? Could it be this: https://docs.rs/syn/latest/syn/struct.ExprUnary.html

ModProg: We could also, if we fancy, evaluate “binop-trees” on integers, as those would probably be the most like used, e.g., `1 << 4` or something like that.

Though, this is a stark complication, as we are basically implementing our own rust expression evaluator at that point.

src/trait_/common_ord.rs
```diff
@@ -214,79 +231,131 @@ pub fn build_ord_signature(
...
-				.map(|(_, variant_other)| variant_other.other_pattern_skip().clone())
-				.collect();
-
-			// Build one match arm pattern with all variants that are `Greater`.
-			let pattern_less = PatOr {
-				attrs: Vec::new(),
-				leading_vert: None,
-				cases,
+					if !has_non_isize {
+						if let Ok(max) = u128::try_from(isize::MAX) {
```
ModProg, line 285: errr... what would happen if an isize doesn't fit a u128, I feel like this would be a place for an expect instead of silently ignoring

src/trait_/common_ord.rs
daxpedda: > that does `-10` parse to? Could it be this: https://docs.rs/syn/latest/syn/struct.ExprUnary.html

I don't know, we will find out after I add the tests.

> We could also, if we fancy, evaluate “binop-trees” on integers, as those would probably be the most like used, e.g., `1 << 4` or something like that.
> 
> Though, this is a stark complication, as we are basically implementing our own rust expression evaluator at that point.

I'm not sure we gain anything because the compiler is supposed to optimize `(1 << 4) + 1` away anyway.

src/trait_/common_ord.rs
daxpedda: I'm not exactly ignoring it, but I assume that it's not an `isize`, which is fine.

But actually I just realized that this whole `isize` check can't work, because I'm not using the target architecture `isize` but the build architecture one.

I will remove it completely ...

daxpedda: Doesn't it say somewhere in the specification the minimum target architecture for Rust has to be 16-bit? I could check for that.

src/lib.rs
daxpedda: Done.

src/trait_/common_ord.rs
daxpedda: Took target architecture pointer width into account.

src/trait_/common_ord.rs
ModProg: Yeah, this is only relevant for cases where this would allow us to use an `isize` instead of needing to fall back to an `i128`.

daxpedda:
Added integration tests, doing unit tests next, but tell me if you think we should add any more specific integration tests.

tests/discriminant.rs
```diff
@@ -1,177 +1,471 @@
...
+fn default_skip() {
+	#[derive_where(PartialEq, PartialOrd)]
 	enum Test {
 		A,
-		B,
+		B = 3,
 		C,
+		#[allow(dead_code)]
+		#[derive_where(incomparable)]
+		D,
```
ModProg, line 117: let's have an `E` after the skip

ModProg (changes requested):

ModProg:
> Added integration tests, doing unit tests next, but tell me if you think we should add any more specific integration tests.

LGTM

tests/discriminant.rs
daxpedda: Done.

src/trait_/common_ord.rs
daxpedda: Confirmed `-10` is parsed as `ExprUnary`.
Not sure if it makes a difference, will add more tests until I see what happens.

daxpedda: I removed all expression parsing now.
The whole logic is much smaller and simplified now, we always fallback to `isize` if we are using the default representation. So this should be solid now.

ModProg: > Confirmed `-10` is parsed as `ExprUnary`. Not sure if it makes a difference, will add more tests until I see what happens.

Well if we only parse numbers `> i16::MIN` we wouldn't need to fall back to `i128`.

daxpedda:
So the magic in the logic we were missing is that:
```rust
// or default
#[repr(C)]
enum Test {
    A(u8) = 5,
    B,
}
```
... doesn't actually work, you can't specify discriminants without specifying the integer type for enums with values. So that way if an enum has values, you ALWAYS know the integer representation. And if it doesn't, you can simply use `var as isize` ... problem solved.

src/item.rs
```diff
@@ -138,7 +137,7 @@ impl Discriminant {
 						}
 					}
 				} else {
-					return Err(Error::repr_parse(attr.span()));
+					unreachable!("found invalid `repr` attribute")
```
ModProg, line 140: did we check, that rustc detects that before executing the proc_macros?

src/trait_/common_ord.rs
daxpedda: I removed that as well, we don't do any parsing at all now.
See https://github.com/ModProg/derive-where/pull/77#issuecomment-1703012002 how this isn't a problem anymore.

ModProg:
> So the magic in the logic we were missing is that:
> 
> ```rust
> // or default
> #[repr(C)]
> enum Test {
>     A(u8) = 5,
>     B,
> }
> ```
> 
> ... doesn't actually work, you can't specify discriminants without specifying the integer type for enums with values. So that way if an enum has values, you ALWAYS know the integer representation. And if it doesn't, you can simply use `var as isize` ... problem solved.

Oh right, the only question would be if, `as isize` is always the correct integer type to use, or if a `repr(C)` could for some reason be a `usize`. Because that is nowhere specified AFAICT.

src/trait_/common_ord.rs
```diff
@@ -214,79 +231,89 @@ pub fn build_ord_signature(
...
+			let discriminant = discriminant.deref();
 
+			quote! {
+				#pattern => #discriminant
+			}
+		});
+
+	// `isize` is currently used by Rust as the default representation when none is
+	// defined.
+	let repr = repr.unwrap_or(Representation::ISize).to_token();
```
ModProg, line 301: This assumption is the only thing I'm not 100% sure is sound

ModProg:
> > Oh right, the only question would be if, `as isize` is always the correct integer type to use, or if a `repr(C)` could for some reason be a `usize`. Because that is nowhere specified AFAICT.
> 
> Yeah it should just cast and it can't overflow because the numbers just start incrementing from 0 upwards. So unless you have a variant with more then 32.767 variants, it's not a problem.

but you can specify other starting points

tests/discriminant.rs
```diff
@@ -1,177 +1,673 @@
...
+fn default_skip() {
+	#[derive_where(PartialEq, PartialOrd)]
 	enum Test {
-		A = 3,
+		A,
+		B = 3,
+		C,
+		#[allow(dead_code)]
+		#[derive_where(incomparable)]
+		E,
```
ModProg, line 119: that is not what I meant with "let's have an E"...

I meant
```suggestion
		#[allow(dead_code)]
		#[derive_where(incomparable)]
		D,
		E,
```

ModProg (changes requested):

src/item.rs
daxpedda: I did, yeah. Our proc-macro isn't even executed in that case.

src/trait_/common_ord.rs
daxpedda: Hm, I have to think about this again, I've found an error in my logic ... again.

tests/discriminant.rs
daxpedda: :sweat_smile:

src/trait_/common_ord.rs
daxpedda: Yeah so the issue is if have an enum with only empty variants, `repr(C)` and custom discriminents, and there is a platform that doesn't use `isize`, we might have an issue.

But I actually have a good answer to that: it's not a problem! Compilation will just fail if you use any funny expressions or literals that are bigger then `isize`. So there is no undefined behavior, it's just that we don't support `repr(C)` unless it has `isize`, which is currently the case on all platforms anyway.

daxpedda:
Man, so many bugs ...

src/trait_/common_ord.rs
ModProg: `as` only fails on debug, it wraps on release right?

daxpedda:
> but you can specify other starting points

No in the cases we were discussing above you can't specify another starting point.
This discussion is getting complex :/

src/trait_/common_ord.rs
daxpedda: I'm loosing my mind a bit because I can't figure out the logic of what I wrote above there.
Still thinking ...

ModProg: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021

assuming, the `repr(u64)` is a `repr(C)` which for some reason behaves like a `repr(u64)`

ModProg:
> No in the cases we were discussing above you can't specify another starting point.

My comment wasn't mean on the enums with fields, because those are a solved problem AFAICT.

The remaining problem are field-less enums. Here one could run into issues of overflowing, in case, for some reason, `isize` isn't the implicit repr. (https://github.com/ModProg/derive-where/pull/77#discussion_r1313300597)

src/trait_/common_ord.rs
daxpedda: We built in some validation for `repr(C)` to make sure it's `isize`.

tests/discriminant.rs
daxpedda: Done.

daxpedda:
I put "Fixes #73" in OP.

tests/ui/discriminant.stderr
```diff
@@ -0,0 +1,19 @@
...
+  --> tests/ui/discriminant.rs:12:1
+   |
+12 | enum MissingIntegerRepr<T> {
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+error[E0732]: `#[repr(inttype)]` must be specified
+  --> tests/ui/discriminant.rs:18:1
+   |
+18 | enum MissingIntegerReprC<T> {
+   | ^^^^^^^^^^^^^^^^^^^^^^^^^^^
```
ModProg, line 19: Interesting, seems like our error isn't even returned...

ModProg (changes requested):

src/test/discriminant.rs
```diff
@@ -0,0 +1,2061 @@
...
+						return ::core::option::Option::None;
+					}
+
+					#discriminant
+
+					if __self_disc == __other_disc {
+						::core::option::Option::Some(::core::cmp::Ordering::Equal)
+					} else {
+						#partial_ord
+					}
```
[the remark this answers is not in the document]
daxpedda: I think this might be better indeed?

tests/ui/discriminant.stderr
daxpedda: Yeah forgot to mention that, it shows up in Rust Analyzer, but not by the compiler, so I'm still in favor to leave our error ... but yeah, no clue.

ModProg: Yeah, let's keep it, especially with the point of a possible future change in rustc

ModProg (approved):

src/test/discriminant.rs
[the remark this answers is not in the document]
daxpedda: Working on it, just need to fix the tests.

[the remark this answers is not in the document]
daxpedda: This was a great catch :heart:!

ModProg (approved):

PR #80, opened 2023-09-02
Improve codegen for enums

daxpedda:
Some more potential improvements I noticed while writing up a report.


ModProg (approved):

ModProg:
With the new information from https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/enum.20discriminant.20size/near/388698648 maybe the check doesn't make sense at all...

daxpedda:
I think we should still leave it until we have more clear guidance by the specification. Because maybe a new target could be introduced that doesn't use `isize` for discriminants without a edition change at all ...

PR #81, opened 2023-09-03
Enforce default discriminant type for all representations and across editions

daxpedda:
See [Zulip](https://rust-lang.zulipchat.com/#narrow/stream/182449-t-compiler.2Fhelp/topic/enum.20discriminant.20size/near/388771801).

Yanking time ... cry me a river ...

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   possible.
 - Avoid unnecessarily validating the discriminant type of enums with C
   representations in some cases.
+- Make sure to validate that enums with C representation use `isize`
+  discriminants even in the case of an edition change.
 
 ## [1.2.4] - 2023-09-01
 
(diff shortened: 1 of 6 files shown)


ModProg:
we could do this for all cases right? even the non `C` one? I feel like that would make sense, as technically also the `repr(Rust)` could change on an edition.

daxpedda:
I was having the same thought, but I would have said that would break all proc-macros out there, so we should be safe? I don't want to further ruin the code generation ...

WDYT?

daxpedda:
We could also ask on Zulip again?

ModProg:
well, I'd say even the change to `repr(C)` would break all proc macros.

src/test/basic.rs
```diff
@@ -197,7 +197,7 @@ fn enum_() -> Result<()> {
 	};
 	#[cfg(not(feature = "nightly"))]
 	let ord = quote! {
-		fn __discriminant<T>(__this: &Test<T>) -> isize {
+		const fn __discriminant<T>(__this: &Test<T>) -> isize {
```
daxpedda, line 200: WDYT of that?

ModProg: Does a const function ensure what we want? i.e. make overflow a compilation error?

ModProg: I checked that it doesn't in #82

daxpedda: Split off to #82.

daxpedda:
It's done, checking both now.

daxpedda:
I think we can get this out first and figure out #83 later when we get an answer.

ModProg (approved):

PR #82, opened 2023-09-03
Make `discriminant` function `const`

daxpedda:
Split off from #81.

### DIFF
--- a/src/test/basic.rs
+++ b/src/test/basic.rs
@@ -197,7 +197,7 @@ fn enum_() -> Result<()> {
 	};
 	#[cfg(not(feature = "nightly"))]
 	let ord = quote! {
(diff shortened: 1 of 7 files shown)


ModProg (approved):
This would only leave the `as` code

ModProg:
```rs
fn main() {
    dbg!(hello());
}

const fn hello() -> u8 {
    200_u8.pow(2)
}
```

runs in release without issues.

While 
```rs
const fn hello() -> u8 {
    const A: u8 = 200_u8.pow(2);
    A
}
```

fails to compile

ModProg:
So while this definitly doesn't hurt, as the function can be const, it doesn't improve the issue at all.

daxpedda:
Oh yeah, it doesn't actually fix our problem, I just added it help the compiler :sweat_smile:.

PR #83, opened 2023-09-03
Remove C validation

daxpedda:
Just leaving this here in case it turns out we don't need this validation ...


daxpedda:
I think we can just go ahead with this. As noted in the discussion back then, Rust can't change this without all hell breaking loose.

ModProg (approved):

PR #87, opened 2023-12-04
Fix wrong usage of `Copy` in `Clone` when bounds are present

### LINKED ISSUE #86: Copy + Clone optimization is incorrect
The code [here](https://github.com/ModProg/derive-where/blob/d0dbebd33a27a7a8863445901004818e32f1088a/src/trait_/clone.rs#L52-L58) that special-cases deriving `Copy` and `Clone` bounds together is incorrect. Take, for example, the following type:

```rust
pub struct Wrapper<Inner, S> {
    inner: Inner,
    marker: PhantomData<S>,
}
```

A `#[derive_where(Copy, Clone; Inner)]` annotation will produce the following code:
```rust
impl<Inner, S> ::core::marker::Copy for Wrapper<Inner, S> where Inner: ::core::marker::Copy {}
impl<Inner, S> ::core::clone::Clone for Wrapper<Inner, S>
where
    Inner: ::core::clone::Clone,
{
    #[inline]
    fn clone(&self) -> Self {
        *self
    }
}
```

This [won't compile](https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=fa6a4dca39f47de6168770fa18727495) because the `Clone` impl can't prove that the type `Inner` implements `Copy` (correctly).

One solution is to require `Inner: Copy + Clone` in the annotation; that produces code that compiles, but that is overly restrictive since `Wrapper<Inner, S>` will only implement `Clone` if `Inner: Clone` _and_ `Inner: Copy`.

daxpedda:
Still going to add some integration tests.

Fixes #86.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,11 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+- Use the `Copy` implementation for `Clone` only if no bounds are present.
+
 ## [1.2.5] - 2023-09-03
 
 ### Changed
--- a/src/test/bound.rs
+++ b/src/test/bound.rs
@@ -152,7 +152,9 @@ fn check_trait_bounds() -> Result<()> {
 			{
 				#[inline]
 				fn clone(&self) -> Self {
-					*self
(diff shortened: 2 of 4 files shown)


ModProg (approved):
How does std handle this? Same logic? (Meaning they only use it without generics)

daxpedda:
I don't know, will check and report back when I'm done with fixing CI.

daxpedda:
As far as I can tell, Rust std doesn't actually use `Copy`, it will always use `clone()` when implementing `Clone`.

PR #93, opened 2023-12-14
Improve CI

daxpedda:
- Don't run CI on every push, but only on the `main` branch and in PRs.
- Cancel CI when a new one is already run on a newer commit.
- Use [`CARGO_TERM_COLOR`](https://doc.rust-lang.org/1.74.0/cargo/reference/environment-variables.html#configuration-environment-variables).
- Use [`taiki-e/install-action`](https://github.com/taiki-e/install-action).
- Make sure to always checkout the repo first, which avoids some issues, like running actions.
- Add [Typos](https://crates.io/crates/typos) for spell checking.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,18 +1,30 @@
 name: Test
 
-on: [push, pull_request]
+on:
+  push:
+    branches: ["main"]
+  pull_request:
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref_name }}
+  cancel-in-progress: true
+
+env:
+  CARGO_TERM_COLOR: always
 
 jobs:
   build-test:
-    name: Build + Test
+    name: Build & Test
+
+    runs-on: ubuntu-latest
 
     strategy:
       fail-fast: false
       matrix:
         rust:
-          - 1.57.0
-          - stable
-          - nightly
+          - { version: 1.57.0, msrv: true }
+          - { version: stable, msrv: false }
+          - { version: nightly, msrv: false }
         features:
           - ""
           - --features safe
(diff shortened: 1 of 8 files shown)


.github/workflows/test.yml
```diff
@@ -1,18 +1,30 @@
...
     strategy:
       fail-fast: false
       matrix:
         rust:
-          - 1.57.0
-          - stable
-          - nightly
+          - { version: 1.57.0, msrv: true }
+          - { version: stable, msrv: false }
+          - { version: nightly, msrv: false }
```
ModProg, line 27: Couldn't you have the msrv on an include?
something like:
```yml
rust: 
  - stable
  - nightly
include:
  - rust: 1.57.0
    msrv: true
```

or
```yml
rust: 
  - stable
  - nightly
  - 1.57.0
include:
  - rust: 1.57.0
    msrv: true
```

This seems to be almost the same usecase: https://docs.github.com/en/actions/using-jobs/using-a-matrix-for-your-jobs#example-expanding-configurations

ModProg (changes requested):

.github/workflows/test.yml
daxpedda: Lets try.

daxpedda: Okay, unfortunately it doesn't seem to work because then it doesn't generate jobs for all the different features in the matrix.

daxpedda: You can see that here: https://github.com/ModProg/derive-where/actions/runs/7208420043.
Ignoring the error (which is just me forgetting to adjust some more stuff in `no_std`), you can see it only generates a single job for v1.57.0.

daxpedda:
Ah, I see, didn't pay close attention.
That works indeed.

ModProg (approved):

PR #95, opened 2023-12-14
Use optimized output for `Clone` and `PartialOrd` with custom bounds

daxpedda:
Some more `Clone` and `PartialOrd` optimization.
Thank you @eric-seppanen for pointing it out.

### DIFF
--- a/src/test/mod.rs
+++ b/src/test/mod.rs
@@ -1,5 +1,6 @@
 mod basic;
 mod bound;
+mod clone;
 mod discriminant;
 mod enum_;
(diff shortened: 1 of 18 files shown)


src/test/mod.rs
```diff
@@ -1,5 +1,6 @@
 mod basic;
 mod bound;
+mod clone;
```
daxpedda, line 3: We somehow managed to forget adding this module.

eric-seppanen:
The macro output looks good in my use case. Thanks!

ModProg (approved):

PR #102, opened 2024-12-01
Fix source file permissions.

alebastr:
`lib.rs` seems to be erroneously marked as executable. This has no purpose, and combined with the `#![...]` in the first line confuses any tools that expect the shebang line.

In particular, our (Fedora) packaging tools consider this source file a script and trip on the invalid shebang while trying to infer the interpreter.


daxpedda (approved):
How did that happen.
Maybe we should consider adding a check in the CI for that?

daxpedda:
CI was fixed in #103.

PR #105, opened 2025-01-30
skip(Clone)

### LINKED ISSUE #104: Feature: Support `skip` on `Clone` derives
Sometimes, it's useful to derive `Clone` for types with fields that cannot be cloned, and need to be re-initialized:

```rust
#[derive_where(Clone, Debug)]
struct Foo {
    // Defaults to `Option::<TcpStream>::default()` on clone
    #[derive_where(skip(Clone))]
    connection: Option<TcpStream>,
    s: String,
    #[derive_where(skip(Debug))]
    i: i32,
}
```

Alternatively maybe `#[derive_where(default(Clone))]` makes more sense here.

ModProg:
fixes #104

### DIFF
--- a/src/trait_/clone.rs
+++ b/src/trait_/clone.rs
@@ -5,7 +5,8 @@ use quote::quote;
 use syn::{TraitBound, TraitBoundModifier, TypeParamBound};
 
 use crate::{
-	Data, DataType, DeriveTrait, DeriveWhere, Item, SimpleType, SplitGenerics, Trait, TraitImpl,
+	data::Field, Data, DataType, DeriveTrait, DeriveWhere, Item, SimpleType, SplitGenerics, Trait,
+	TraitImpl,
 };
 
 /// Dummy-struct implement [`Trait`] for [`Clone`](trait@std::clone::Clone).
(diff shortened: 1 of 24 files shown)


daxpedda (commented):
Just hypothetically, what will we do when we get to a breaking change? Will we ever want `derive_where(skip)` to make `Clone` just call `default()` instead?

I'm thinking that letting `Clone` call `default()` under the term `skip` is not ideal and would clash with the general usage of `derive_where(skip)`. I propose instead a new attribute: `derive_where(clone_default)` or `derive_where(clone(default))`. WDYT?

Will make a review of the code properly after figuring this out, because it would change the code entirely probably.

ModProg:
I think I'm getting to a point there I'm questioning the global skip as a feature. I think it makes sense to have skips for EqHash etc. but just skipping for all traits *that currently support the feature* seems strange to me now

daxpedda:
> I think I'm getting to a point there I'm questioning the global skip as a feature. I think it makes sense to have skips for EqHash etc. but just skipping for all traits _that currently support the feature_ seems strange to me now

Didn't cross my mind at all, but seems quite reasonable to me. In hindsight, a global skip almost looks like a footgun.

But unrelated, right? Or did you mean to keep `skip(Clone)` instead of a dedicated attribute considering you want to remove the global skip in the future?

ModProg:
I feel like skip(Clone) is in line with other things that don't explicitly specify that `Default` will be used, when "removing" a value e.g. https://doc.rust-lang.org/std/mem/fn.take.html

daxpedda:
Hm, `skip` to me implies that *nothing* is done, like with all the other traits `skip` supports right now. `Clone` definitely seems like a departure to me in this regard. While `default()` can imply something like "empty", as with `std::mem::take()`, it seems still different to me then "nothing" as "not used" or "skipped".

I will go ahead and review nevertheless, not a big issue really.

ModProg:
I see your point, but thought of skip more as *not doing the specific thing* in this case cloning, but still doing what is required to allow the skipping, in this case defaulting.

src/trait_/clone.rs
```diff
@@ -99,25 +100,26 @@ impl TraitImpl for Clone {
 		}
 
 		match data.simple_type() {
-			SimpleType::Struct(fields) => {
+			SimpleType::Struct(fields) | SimpleType::Tuple(fields) => {
```
daxpedda, line 103: We might be able to simplify some more code with that notation.

daxpedda (changes requested):
Looks great!
Just needs an update of the README.

src/trait_/clone.rs
daxpedda: Ofc! Was just noting :)

daxpedda (approved):

ModProg:
@daxpedda we need to mention that `skip` without group does not include clone

ModProg:
maybe referencing #111

daxpedda:
I specifically did *not* include `Clone` in  the list for blanket skipping, but only in the list for group skips.
Doesn't that cover it?

ModProg:
True, should be fine, especially with the aforementioned planed deprecation of the "skip all" feature.

PR #107, opened 2025-04-13
Feat: Allow lifetime for bounds in non-binded generic params

stegaBOB:
It's currently not possible to use higher rank trait bounds without providing custom predicate type bounds. Our particular use case is enabling "trivial bounds" without requiring the nightly feature. For context: https://github.com/rust-lang/rust/issues/48214#issuecomment-2799836786

For example:

```rust
// fails due to "the trait bound `std::string::String: std::marker::Copy` is not satisfied"
#[derive_where(Clone, Copy; String)]
struct TestThing {
    a: String,
}

// Compiles! Copy is just not implemented due to String obviously not being copy, so it would compile error on use.
#[derive_where(Clone, Copy; for<'a> String)]
struct TestThing {
    a: String,
}
```

This is useful for adding perfect derives to macro-generated structs. 

I'm not fully sure whether having a HRTB should make `has_type_param` return false or not. I think its probably fine to keep it returning true since adding one that's not used doesn't change the behavior; it should still be derivable w/ the standard derive. Users can always make type aliases to get around this anyways if they disagree lol.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+
 ### Added
 - Allow lifetime `for<'a, ...>` bounds in non-bounded generic parameters.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
+- Allow lifetime `for<'a, ...>` bounds in non-bounded generic parameters.
 
 ### Changed
 - Use the `Copy` implementation for `Clone` and the `Ord` implementation for
--- a/src/attr/item.rs
+++ b/src/attr/item.rs
@@ -241,10 +241,10 @@ impl DeriveWhere {
 	/// Returns `true` if the given generic type parameter if present.
 	pub fn has_type_param(&self, type_param: &Ident) -> bool {
 		self.generics.iter().any(|generic| match generic {
-			Generic::NoBound(GenericNoBound(
-				_lifetimes,
-				Type::Path(TypePath { qself: None, path }),
-			)) => {
+			Generic::NoBound(GenericNoBound {
+				lifetimes: _,
+				ty: Type::Path(TypePath { qself: None, path }),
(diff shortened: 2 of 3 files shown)


src/attr/item.rs
```diff
@@ -293,21 +296,31 @@ impl DeriveWhere {
 	}
 }
 
-/// Holds a single generic [type](Type) or [type with bound](PredicateType).
+/// Holds the first part of a [`PredicateType`] prior to the `:`. Optionally contains lifetime `for`
+/// bindings.
+#[derive(Eq, PartialEq)]
+pub struct GenericNoBound(Option<BoundLifetimes>, Type);
+impl Parse for GenericNoBound {
```
daxpedda, line 303:
```suggestion
#[derive(Eq, PartialEq)]
pub struct GenericNoBound(Option<BoundLifetimes>, Type);

impl Parse for GenericNoBound {
```

src/attr/item.rs
```diff
@@ -293,21 +296,31 @@ impl DeriveWhere {
 	}
 }
 
-/// Holds a single generic [type](Type) or [type with bound](PredicateType).
+/// Holds the first part of a [`PredicateType`] prior to the `:`. Optionally contains lifetime `for`
+/// bindings.
+#[derive(Eq, PartialEq)]
+pub struct GenericNoBound(Option<BoundLifetimes>, Type);
```
daxpedda, line 302: Lets use named fields here.

daxpedda (changes requested):
This is indeed a very strange quirk of Rust!

Please add a changelog entry as well.
Otherwise LGTM!

src/attr/item.rs
stegaBOB: Done!

daxpedda (changes requested):
- Run `cargo +nightly fmt`.
- Changelog entry.

stegaBOB:
Completely missed the changelog request! Sorry about that. Should be good to go now I think

CHANGELOG.md
```diff
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
+### Added
```
daxpedda, line 9:
```suggestion
## [Unreleased]

### Added
```

daxpedda (approved):
> Completely missed the changelog request! Sorry about that. 

All good! I know how it is :)

Will wait for #108 before merging.

stegaBOB:
Hey @daxpedda was wondering if there are any plans to cut a new release soon. We're currently working on a private registry so we aren't blocked on this or anything, but just wanted to ask since its been a while since the last one. Thanks! :)

PR #109, opened 2025-04-14
Support skipping only some variants with `ZeroizeOnDrop`

daxpedda:
I noticed that the following would fail to compile:
```rust
#[derive_where(ZeroizeOnDrop)]
enum Test<T> {
    A(PhantomData<T>),
    #[derive_where(skip_inner)]
    B(PhantomData<T>),
}
```
Because it produces the following:
```rust
impl<T> Drop for Test<T> {
    fn drop(&mut self) {
        match self {
            Test::A(ref mut __field_0) => {
                __field_0.zeroize_or_on_drop();
            }
        }
    }
}
```
Instead, we now make sure to also cover variants that are skipped, but their body is just empty now.

See the test I added for a more complete example.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,11 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Use the `Copy` implementation for `Clone` and the `Ord` implementation for
   `PartialOrd` when custom bounds are present.
 
+### Fixed
+- Support skipping only some variants with `ZeroizeOnDrop`.
+- Only generate single call to `zeroize()` with `ZeroizeOnDrop` on multiple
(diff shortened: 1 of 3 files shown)


daxpedda:
I discovered another bug!

Apparently the following:
```rust
#[derive_where(ZeroizeOnDrop)]
enum Test<T> {
    A(PhantomData<T>),
    #[derive_where(skip_inner)]
    B(PhantomData<T>),
}
```
Would generate multiple calls to `Zeroize::zeroize()`:
```rust
impl<T> Drop for Test<T> {
    fn drop(&mut self) {
        Zeroize::zeroize(self);
        Zeroize::zeroize(self);
    }
}
```

ModProg (approved):

PR #114, opened 2025-05-09
Implement `ZeroizeOnDrop` without `Drop`

daxpedda:
This PR adds a new option to `ZeroizeOnDrop`: `no_drop`. This will implement `ZeroizeOnDrop` without implementing `Drop`. Instead, it will assert that all fields implement `ZeroizeOnDrop` (not `Zeroize`!).

The second part of this PR reworks `DeriveTrait` and moves the ownership of derive related data to the actual `TraitImpl` instead of it just being a dummy-struct to implement `TraitImpl` over. Now we have to pass in less parameters to each method and can actually use `self` to store data.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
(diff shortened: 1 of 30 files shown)


ModProg:
Wondering if we could have the same behavior on field level? Though that could require a better name.

daxpedda:
> Wondering if we could have the same behavior on field level? Though that could require a better name.

That is the already the default, in `Drop` it will detect if it implements `Zeroize` **or** `ZeroizeOnDrop`.

ModProg (approved):

PR #116, opened 2025-05-14
Stop Using Internal APIs

### LINKED ISSUE #115: Don't use internal APIs
We currently have two uses of internal API:
- [`zeroize::__internal`](https://github.com/ModProg/derive-where/blob/4980b4ed4e7f6c113714c91f932f141b0035ab3a/src/test/zeroize.rs#L55-L56).
- [`Eq::assert_receiver_is_total_eq()`](https://github.com/ModProg/derive-where/blob/4980b4ed4e7f6c113714c91f932f141b0035ab3a/src/test/basic.rs#L51-L56).

Both are `doc(hidden)`, so they should be considered unstable. Instead we just have to copy the implementation and emit it ourselves.

daxpedda:
This changes the implementation of `Eq` to not use `Eq::assert_receiver_is_total_eq()` and the implementation of `ZeroizeOnDrop` to not use `zeroite::__internal`. Both are `doc(hidden)` and therefor unstable APIs.

Resolves #115.

### DIFF
--- a/src/test/basic.rs
+++ b/src/test/basic.rs
@@ -45,16 +45,23 @@ fn struct_() -> Result<()> {
 				}
 			}
 
-			#[automatically_derived]
-			impl<T> ::core::cmp::Eq for Test<T> {
-				#[inline]
-				fn assert_receiver_is_total_eq(&self) {
-					struct __AssertEq<__T: ::core::cmp::Eq + ?::core::marker::Sized>(::core::marker::PhantomData<__T>);
-
-					// For some reason the comparison fails without the extra space at the end.
-					let _: __AssertEq<std::marker::PhantomData<T> >;
+			const _: () = {
+				trait DeriveWhereAssertEq {
+					fn assert(&self);
 				}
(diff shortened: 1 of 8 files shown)


src/test/basic.rs
```diff
@@ -45,16 +45,23 @@ fn struct_() -> Result<()> {
...
+			const _: () = {
+				trait DeriveWhereAssertEq {
+					fn assert(&self);
+				}
 
-					// For some reason the comparison fails without the extra space at the end.
-					let _: __AssertEq<std::marker::PhantomData<T> >;
+				impl<T> DeriveWhereAssertEq for Test<T> {
+					fn assert(&self) {
+						struct __AssertEq<__T: ::core::cmp::Eq + ?::core::marker::Sized>(::core::marker::PhantomData<__T>);
```
ModProg, line 55: at some point we could consider creating our own types for stuff like that, though that'd require a non procmacro crate

ModProg (approved):

PR #118, opened 2025-05-15
Add Serde Support

### LINKED ISSUE #100: `derive_where` for serde
Would it be possible to support serde derive macros (`Serialize` & `Deserialize`) in `derive-where`?

Currently, if you put fields from associated types, you have to manually put `#[serde(bounds(serialize = "XXX: Serialize", deserialize = "XXX: Deserialize<'de>"))]`, which can amount to a lot of boilerplate, and might not even be possible to generate in declarative macros (at least I didn't find any way to).

To not reimplement the whole `serde` in this crate, you can just emit the corresponding `#[serde(bounds)]` and hand off the rest to the upstream derive macro

daxpedda:
Adds `serde::De/Serialize` support.

I've got the idea from https://github.com/serde-rs/serde/issues/1900. The idea is to basically emit the whole item in the attribute macro again, let `serde_derive` do its thing and then remove it with another attribute macro that just voids the duplicate item again.

Resolves #100.


daxpedda:
The implementation is done, barring any bugs that still come up during testing.
I'm gonna add a complete test suite later today.

daxpedda:
This is good to go!

ModProg (approved):

fjarri:
This is great! Could you make a release with this change?

PR #128, opened 2025-08-09
don't generate extra `impl` for `Eq` assertions

fee1-dead:
`Eq` already has a method `assert_receiver_is_total_eq` that allows you to put asserts in. This PR makes it so we use that instead of generating a new trait and impl every time (which could be a lot).

This is also a cleanup because apparently `additional_impl` is only used by `Eq` (and arguably quite confusing as `additional_impl` is used to generate the `Eq` impl while the "main" `impl` was the extra impl) so we can remove that now.

generating a `struct __AssertEq` each time is still a little wasteful, but given that `derive-where` is a single crate (so no "normal" support crate to share types) it might be unavoidable.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -648,20 +648,7 @@ fn generate_impl(
 	let body = generate_body(derive_where, trait_, item, generics);
 
 	let ident = item.ident();
-	let mut output = trait_.impl_item(crate_, full_item, imp, ident, ty, &where_clause, body);
-
-	if let Some((path, body)) = trait_.additional_impl() {
-		output.extend(quote! {
-			#[automatically_derived]
-			impl #imp #path for #ident #ty
-			#where_clause
-			{
-				#body
-			}
-		})
-	}
-
-	output
+	trait_.impl_item(crate_, full_item, imp, ident, ty, &where_clause, body)
 }
 
 /// Generate implementation method body for a [`Trait`].
--- a/src/test/basic.rs
+++ b/src/test/basic.rs
@@ -45,23 +45,15 @@ fn struct_() -> Result<()> {
 				}
 			}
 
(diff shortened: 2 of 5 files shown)


daxpedda:
All methods in Std marked with `#[doc(hidden)]` are *API* unstable. `#[stable]` just means its available on Rust stable, it doesn't mean it is *API* stable. E.g. I remember there being a `#[doc(hidden)]` method to define where the output of `println!` goes that was simply removed.

I'm happy to be told otherwise if you can find me something concrete on that topic, I myself don't have anything concrete on it.

daxpedda:
> `set_print` had an `#[unstable]` attribute, i.e. you had to be on nightly and use `feature()` to use it:

Ah, good catch.

> So
> 
> > `#[stable]` just means its available on Rust stable, it doesn't mean it is _API_ stable.
> 
> is not true. Rust stdlib maintainers cannot make breaking changes to stable code. That's outlined in https://rust-lang.github.io/rfcs/1105-api-evolution.html - `#[stable]` absolutely constitutes stable code.

I'm not seeing that `#[stable]` constitutes stable API in RFC 1105. Can you quote the relevant paragraph?

daxpedda:
Worst case scenario, I'm happy to take a quote from a Rust team member on e.g. Zulip or something.

fee1-dead:
See also https://rustc-dev-guide.rust-lang.org/stability.html.

I'm a Rust team member, although I primary work with the compiler (I'm a compiler maintainer). I am confident that `#[stable]` means `assert_receiver_is_total_eq` will never get removed.

fee1-dead:
> Apparently it is considered unstable.

Oh well. That part is really unintuitive. Thanks for asking! I stand corrected

PR #130, opened 2025-08-12
pin glob for msrv

ModProg:


daxpedda (changes requested):
Can we do this conditionally on the MSRV only?

ModProg:
we do, look up 4 lines

PR #143, opened 2026-08-11
Update alternative section

197g:
There's a RUSTSEC unmaintained notice for `derivative` ([RUSTSEC-2024-0388](https://rustsec.org/advisories/RUSTSEC-2024-0388.html)). The new crate added to the alternatives section is one of the options also referenced in this note.


ModProg (approved):
Thank you

PR #144, opened 2026-09-12
update to syn v3

ModProg:


src/lib.rs
daxpedda: Needs to be synced with the README.

Unfortunately #31 didn't pan out yet.

daxpedda (approved):
Well, that was easy.

src/lib.rs
ModProg: good catch

